PR #2, opened 2020-07-23
Update README.md

mattvenn:
made it clear where to clone the open_pdks repo


agorararmard:
@mattvenn can you please modify it to: cd ../pdks instead of cd pdks
Because we are working on the assumption that the user will follow the previous instructions as well, and this way he should already be inside pdks.

Also please submit the PR to the develop branch.

PR #32, opened 2020-09-04
Develop antenna or

agorararmard:
Integration of OR ARC controlled by $::env(USE_ARC_ANTENNA_CHECK)


flow.tcl
```diff
@@ -37,10 +37,11 @@ proc run_non_interactive_mode {args} {
 	run_routing
 
 	if { $::env(DIODE_INSERTION_STRATEGY) == 2 } {
-	    run_magic_antenna_check; # produces a report of violators; extraction!
+		run_antenna_check
 	    heal_antenna_violators; # modifies the routed DEF
 	}
-
+	run_antenna_check
```
riking, line 43: This isn't inside an else block, so you're running it twice

Maybe make a comment that this is intentional?

scripts/tcl_commands/all.tcl
```diff
@@ -597,40 +597,56 @@ proc save_views {args} {
...
+	if { $::env(DIODE_INSERTION_STRATEGY) == 2 } {
+		#Legacy from magic antenna checker
+		#set report_file [open $::env(magic_report_file_tag).antenna_violators.rpt r]
+		#	set violators [split [string trim [read $report_file]]]
+		#close $report_file
+		# may need to speed this up for extremely huge files using hash tables
+		#exec echo $violators >> $::env(TMP_DIR)/vios.txt
+		
+		
+		# Shell solution for replacement
```
riking, line 613: This really doesn't need to be kept around here, does it? It can live either in the git history or in the new python scripts as a reference implementation

flow.tcl
agorararmard: The process goes as follows:
if Diode strategy 2 -> run the antenna checks generate the antenna report -> use the antenna report to replace fake diodes with real diodes (reducing the number of antenna violations)

Then, either way, run the antenna checks to generate the final report.

scripts/tcl_commands/all.tcl
agorararmard: I agree. I guess a little clean-up wouldn't hurt and will make the block "less" confusing.

PR #43, opened 2020-09-29
Typo fix in doc

tgingold:
Minor typo fix as the wrong name is deprecated.


agorararmard:
@tgingold, we're well aware of this typo. But, the variable name in the open_pdks used currently has this typo. So, we're waiting for updating it there along with some other stuff and then we'll correct that typo here as well. 
Thank you!

PR #55, opened 2020-10-03
spef_extractor: remove some unused variables and scalarize some others

tgingold:
The minCap/maxCap variables were never really used.
The edgeCapFactor, capCounter and resCounter variables were lists with
always 1 element, so they can be scalarized.


ax3ghazy:
Got it; merged. Thank you.

PR #61, opened 2020-10-07
Develop repair antenna iters

agorararmard:


scripts/tcl_commands/routing.tcl
```diff
@@ -26,16 +26,43 @@ proc global_routing {args} {
...
-      set ::env(DIODE_INSETION_STRATEGY) 3
+        set iter 2
+        set prevDEF1 $::env(SAVE_DEF)
+	set prevDEF2 $::env(SAVE_DEF)
+	set prevAntennaVal [exec grep "#Antenna violations:" $::env(fastroute_log_file_tag).log -s | tail -1 | sed -r "s/.*\[^0-9\]//"]
+        set_def $::env(SAVE_DEF)
+	while {$iter <= $::env(GLB_RT_MAX_DIODE_INS_ITERS) && $prevAntennaVal > 0} {
+            set ::env(SAVE_DEF) $::env(fastroute_tmp_file_tag)_$iter.def
+            set replaceWith "INSDIODE$iter"
+            try_catch sed -i -e "s/ANTENNA/$replaceWith/g" $::env(CURRENT_DEF)
```
ax3ghazy, line 39: If the cell name is "ANTENNA" in the technology, which I believe is the case for some other technology I came across, this will fail.

scripts/tcl_commands/routing.tcl
```diff
@@ -26,16 +26,43 @@ proc global_routing {args} {
...
+            puts_info "FastRoute Iteration $iter"
+            puts_info "Antenna Violations Previous: $prevAntennaVal"
+            try_catch openroad -exit $::env(SCRIPTS_DIR)/openroad/or_route.tcl |& tee $::env(TERMINAL_OUTPUT) $::env(fastroute_log_file_tag)_$iter.log   
+            set currAntennaVal [exec grep "#Antenna violations:"  $::env(fastroute_log_file_tag)_$iter.log -s | tail -1 | sed -r "s/.*\[^0-9\]//"]
+            puts_info "Antenna Violations Current: $currAntennaVal"
+            if { $currAntennaVal >= $prevAntennaVal } {
+                set iter [expr $iter - 1]
```
ax3ghazy, line 46: Does it really happen that the results get worse from an iteration to the next?

scripts/tcl_commands/routing.tcl
```diff
@@ -26,16 +26,43 @@ proc global_routing {args} {
...
+            } else {
+                set prevAntennaVal $currAntennaVal
+                set iter [expr $iter + 1]
+                set prevDEF1 $prevDEF2
+ 	        set prevDEF2 $::env(SAVE_DEF) 
+            }
+	    set_def $::env(SAVE_DEF)
+        }
+        set ::env(DIODE_INSERTION_STRATEGY) 0
+        try_catch openroad -exit $::env(SCRIPTS_DIR)/openroad/or_route.tcl |& tee $::env(TERMINAL_OUTPUT) $::env(fastroute_log_file_tag)_post_antenna.log
```
ax3ghazy, line 58: This will be removed once the antenna flow works properly, correct?

agorararmard: Yes

scripts/tcl_commands/routing.tcl
agorararmard: Yeah, especially in smaller designs, it diverges drastically. for example xtea:
```
iter 0: 88 vios
iter 1: 7 vios
iter 2: 9 vios
iter 3: 15 vios
... and so on
```
So better be safe, and abort when diverging to the latest stable point.

scripts/tcl_commands/routing.tcl
agorararmard: I'll modify it to (`ANTENNA_` to `INSDIODE$iter_`), the problem is that FastRoute uses the keyword `ANTENNA_idx` for insertion so from one iteration to the next 2 cells will be called `ANTENNA_0`, for example. That's why I'm renaming them.

ax3ghazy: That doesn't solve the problem; you need make sure you're changing the instance name not the instance type from within the COMPONENTS section only. (e.g., `- instance_name instance_type....`). If a net or a pin for some reason has the string "ANTENNA_" within, it will inadvertently change.

agorararmard: Roger that.

agorararmard: bbc2de3b0cc6f79451d40909717abe3054f4be05

PR #69, opened 2020-10-12
Add 151 (UC Berkeley CS 151/251A class) design

growly:
This is a RISC-V core written for a class. It is what it is.


ax3ghazy:
That's great. Thanks for sharing it! Now, nothing should be marked as a release or go to `master` unless it successfully hardens your design.

PR #77, opened 2020-10-16
add s44 LUT design

growly:
This is a tiny design, and it doesn't work, but we'd very much like it to.


designs/s44/src/lut.v
```diff
@@ -0,0 +1,43 @@
...
+// Authors: Ryan Thornton (rpthornton@berkeley.edu)
+//          Arya Reais-Parsi (aryap@berkeley.edu)
+
+
+///////// BASIC LUT /////////
+// Assumptions:
+//  MEM_SIZE is a multiple of CONFIG_WIDTH 
+
+module lut #(
+    parameter INPUTS=4, MEM_SIZE=2^INPUTS, CONFIG_WIDTH=8
```
tgingold, line 11: `^` is xor, I suppose you mean: `MEM_SIZE = 1 << INPUTS`
Is there any reason why the user would change `MEM_SIZE` ?  If no, it should be a localparam.

agorararmard (changes requested):
@growly, You should replace the configurations in config.tcl with the following: (These worked for me on the latest openlane develop)
```tcl
# User config
set ::env(DESIGN_NAME) lut_s44

# Design config
set ::env(CLOCK_PERIOD) 30
set ::env(VERILOG_FILES) [glob $::env(OPENLANE_ROOT)/designs/s44/src/*.v]
set ::env(CLOCK_PORT) "config_clk"
set ::env(CLOCK_NET) $::env(CLOCK_PORT)
# Synthesis config
set ::env(SYNTH_STRATEGY) 1
# Floorplan config
set ::env(FP_CORE_UTIL) 5
# Placement config
set ::env(PL_TARGET_DENSITY) 0.5

set ::env(PDN_CFG) $::env(DESIGN_DIR)/pdn.tcl


set filename $::env(DESIGN_DIR)/$::env(PDK)_$::env(STD_CELL_LIBRARY)_config.tcl
if { [file exists $filename] == 1} {
	source $filename
}
```

designs/s44/src/lut.v
growly: good catch on the wrong operator; I think the original author intended it to be an external parameter to enable architecture exploration, so I'll leave it there.

agorararmard (approved):

PR #80, opened 2020-10-30
Develop

growly:
This is the design I've been bugging you with. I just want to put a single macro in the bottom left of an otherwise empty design.


agorararmard:
@growly: Since the design now runs successfully, and we already have a test case that covers this scenario, I'll close this PR.

PR #89, opened 2020-11-14
scripts/libtrim.pl: Don't warn on experimental

smunaut:
Ubuntu uses a perl version that throws a warning. This doesn't stop it
from running but the stderr output is interpreted as an error by the flow
and stopping it. So disable those warnings.

Signed-off-by: Sylvain Munaut <tnt@246tNt.com>

### DIFF
--- a/scripts/libtrim.pl
+++ b/scripts/libtrim.pl
@@ -21,6 +21,8 @@ use strict;
 use feature "switch";
 #use experimental qw( switch );
 
+no warnings 'experimental';
+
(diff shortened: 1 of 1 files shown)


smunaut:
Looks like this breaks on perl version that don't have 'experimental' and got no idea how to fix that so I guess I'll keep that patch to my local tree.

agorararmard:
@smunaut: I was going to mention the same thing. This would break the version in the OpenLANE docker. However, I think it would be a good idea to mention that in the [FAQs](https://github.com/efabless/openlane/wiki) maybe?

PR #92, opened 2020-11-16
CVC: Incorporated CVC into openlane/flow.tcl

d-m-bailey:
This change enables CVC to run on the abstract netlist extracted from `maglef` cells. Currently, CVC on device level extracted netlists is not supported.

1) Create the appropriate CVC directories in `scripts/tcl_commands/all.tcl`
2) Copy the abstract spice file to `*.lef.spice` (change to `run_magic_export_spice` in `scripts/tcl_commands/magic.tcl`
3) Added `run_lef_cvc` to the `scripts/tcl_commands/lvs.tcl` (might be better to add to a separate file, but I didn't know how to setup the tcl library)
4) Call `run_lef_cvc` from `flow.tcl`
5) `run_lef_cvc` creates a power definition file from the verilog file used for LVS and merges the cdl library with the `*lef.spice` file.
6) Model definitions are in `scripts/cvc.sky130.models`
7) CVC parameters are in `scripts/cvcrc.sky130`

Probably want to add error reporting. From the CVC log file the `CVC: Total: `line shows the number of errors.
`CVC: Total:                 0`
Digital design will most likely have no false errors. Analog designs may have false errors that we can deal with later.


agorararmard (approved):

agorararmard:
@d-m-bailey: I will merge that right away since we're using that branch (develop-cvc) for development. However, to give you a heads-up, I'll be replacing the hardcoded power pins with the appropriate environment variables. I'll move the sky130 setup files to [open_pdks](https://github.com/RTimothyEdwards/open_pdks). I'll also add an enable-disable flag for the user's convience. 

However, I don't think there is a lot do from here except more and more testing before this branch is finally merged to develop and then later to master. Great job!

PR #99, opened 2020-11-21
Misc Improvements discussed on Slack

smunaut:


agorararmard (approved):

scripts/tcl_commands/floorplan.tcl
```diff
@@ -216,11 +216,12 @@ proc run_floorplan {args} {
 		}
 
 		if { [info exist ::env(EXTRA_LEFS)] } {
-			global_placement_or
 			if { [info exist ::env(MACRO_PLACEMENT_CFG)] } {
 				file copy -force $::env(MACRO_PLACEMENT_CFG) $::env(TMP_DIR)/macro_placement.cfg
 				manual_macro_placement f
+				global_placement_or
```
ax3ghazy, line 222: @smunaut: Can we do without this line then? The actual placement will happen at a later stage anyway (during `run_placement`), or are you saying that doing `global_placement` *two* times with macros fixed gives better results?

smunaut: I haven't tried removing the line all together ... didn't even occur to me.
I just found it weird to run an auto-placer before fixing what's fixed, so I just swapped them.

ax3ghazy: Yea, it's definitely weird if you're manually placing macros, but it's needed if you're using TritonMP, which requires an initial solution to work with. Anyway, I will merge this and remove this step.

ax3ghazy:
Thanks!

PR #110, opened 2020-11-27
Supress errors from openROAD about missing config files

thesourcerer8:
openROAD complains about those config files missing when running sta and openroad, adding them empty removes those confusing errors.

### DIFF
--- a/docker_build/Dockerfile
+++ b/docker_build/Dockerfile
@@ -69,6 +69,9 @@ ADD ./openroad_tools.tar.gz /
 COPY ./.tclshrc /
 COPY ./.tclshrc /root
 
+RUN touch /.sta
+RUN touch /.openroad
+
 ADD ./openLANE_flow.tar.gz $OPENLANE_ROOT
 
 # install opendb python


agorararmard:
@thesourcerer8: I think it would be wise to wait for their response on this since you already opened an issue there https://github.com/The-OpenROAD-Project/OpenROAD/issues/533 . In case adding the empty files would introduce some undesired behavior.

thesourcerer8:
I dont think that this has any undesired behaviour, they are just trying to load user-specific configurations from ~/.sta like many other unix tools that read both global configs from /etc/apprc and from the users homedirectory ~/.apprc so that the system administrator can define global configs, and that users can override them. Putting an empty file there without any options does not cause problems.

agorararmard:
@thesourcerer8: Thank you very much for your PR, however since this issue was resolved in the OpenROAD master branch, which we'll be using in develop very soon (it's under testing [here](https://github.com/efabless/openlane/tree/develop-or_22122020)), we won't be able to merge your PR.

PR #133, opened 2020-12-20
Fixes for mac

robtaylor:
Some tweaks for getting Openlane start of day experience better on my macbook..


smunaut:
Why do all theses tcl file neet the `x` bit set anyway ?

I run natively (not with docker), never changes the `x` bit on those and I never had any issues ... AFAIK you never call theses directly so it's not needed.

agorararmard:
@robtaylor: Thank you very much for your Pull Request. Could you rebase that to the develop branch in order for us to go forward with it?
https://github.com/efabless/openlane/blob/master/CONTRIBUTING.md

robtaylor:
@agorararmard Thanks, i'd missed that. rebased against develop now.

agorararmard (approved):

robtaylor:
Thank you Amr!

PR #142, opened 2020-12-30
Apply CELL_PAD to replace

antonblanchard:
replace has a somewhat recent option to specify padding.

### DIFF
--- a/scripts/openroad/or_replace.tcl
+++ b/scripts/openroad/or_replace.tcl
@@ -48,6 +48,8 @@ set_replace_verbose_level_cmd 1
 
 set_replace_density_cmd $::env(PL_TARGET_DENSITY)
 
+set_replace_pad_right_cmd $::env(CELL_PAD)
+
 if { $::env(PL_BASIC_PLACEMENT) } {
 	set_replace_overflow_cmd 0.9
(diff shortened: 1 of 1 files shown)


antonblanchard:
I notice a few other differences vs the openroad regression tests:

openroad applies a left and right pad, and openlane applies only one (left for cts, right otherwise). This might make a small improvement to routability I think. Could we apply CELL_PAD/2 left and right?

openroad also applies 2x padding in global placement vs detailed placement, I'm not sure why.

antonblanchard:
I wonder if this will cause some working configurations to fail during global placement

antonblanchard:
It does seem to cause previously passing configurations to fail.

agorararmard:
@antonblanchard: We set these values with experimentation over the test set we provide in the repo: https://github.com/efabless/openlane/tree/master/designs I don't think it will make a huge difference if the padding is to the left or to the right or divided over both. However, if you have a test case that proves the significant difference we'll be happy to reconsider.

It seems that replace's cell padding didn't work well with the current designs and flow setup (maybe it needs to be 2x as you said?). Perhaps we can look into this later. But, for now, I will not be able to merge this PR (check the CI for more details).

antonblanchard:
Agreed, lets close it for now and we can try tuning it sometime later.

PR #143, opened 2021-01-01
Move add_route_obs before global routing

antonblanchard:
On a design with many large macros, FastRoute is placing guides right
through the middle of those macros. TritonRoute never manages to recover
from this and I end up with over a thousand DRC issues due to these tracks.

In order to stop this I added an GLB_RT_OBS line covering all the macros,
but that didn't help FastRoute because the obstructions get added after global
routing. Moving add_route_obs earlier fixes this.

### DIFF
--- a/scripts/tcl_commands/routing.tcl
+++ b/scripts/tcl_commands/routing.tcl
@@ -253,6 +253,8 @@ proc run_routing {args} {
     }
     use_original_lefs
 
+    add_route_obs
+
     global_routing
 
     # insert fill_cells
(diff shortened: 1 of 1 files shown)


smunaut:
Mmm .. that's weird,when I created GLB_RT_OBS, I specifically put it after FastRoute because it would fail if I placed it before ...

agorararmard (approved):

agorararmard:
@antonblanchard: Merged. Thanks!

antonblanchard:
@smunaut I haven't hit any failures but I just noticed one of my macros which previously passed DRC easily no longer does. I have to give it a chunk more area to get it to pass

PR #145, opened 2021-01-02
[opendp] Restore diamond_search_height_ to 400

antonblanchard:
diamond_search_height_ was originally 400 but has since been changed
to 100. On a design with large macros, the openroad version of opendp
fails but the standalone version works.

The upstream fix to openroad is still being discussed in:

https://github.com/The-OpenROAD-Project/OpenROAD/pull/549

but since it is causing issues with at least one caravel design, restore
the old behaviour until the ultimate fix makes it in.

### DIFF
--- a/docker_build/docker/openroad_app/Dockerfile
+++ b/docker_build/docker/openroad_app/Dockerfile
@@ -131,6 +131,9 @@ RUN git apply pdngen_export_subst.patch
 COPY ignore_obs_outside.patch /OpenROAD_22122020/
 RUN git apply ignore_obs_outside.patch
 
+COPY opendp-diamond-search.patch /OpenROAD_22122020/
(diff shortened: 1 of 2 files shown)


agorararmard:
@antonblanchard: Can you also push the resulting tarball here? the result of running:
```bash
cd docker_build/
make build-openroad_app
```

antonblanchard:
I pushed it, but haven't had a chance to test. Does the CI use the tarballs to run the tests?

agorararmard (approved):

agorararmard:
@antonblanchard: Yes, the CI uses the existing tarballs, so it tested your update against these designs: https://travis-ci.com/github/efabless/openlane/builds/211992254 and it passed. So, I'm going to go ahead and merge this PR. Thanks!

PR #147, opened 2021-01-02
Build netgen with optimisation

antonblanchard:
netgen is taking ages to run on my design and I notice it is being built
without any optimisation. A quick test on ppc64le Linux shows -O2 is
almost twice as fast. x86-64 may not gain as much, but it will almost
certainly improve too.

### DIFF
--- a/docker_build/docker/netgen/Dockerfile
+++ b/docker_build/docker/netgen/Dockerfile
@@ -32,7 +32,7 @@ RUN git checkout 920c6e69283f9bef5d3aa67776b868696e438c2d
 
 
 # build
-RUN ./configure --prefix=/build
+RUN ./configure CFLAGS="-O2 -g" --prefix=/build
 
 RUN make clean && \
     make && \


ax3ghazy:
This shouldn't hurt (unless some piece of code relies on an undefined behaviour somewhere in netgen?), but I will merge this for now to develop. Thanks!

d-m-bailey:
@ax3ghazy @antonblanchard Usually the fix given should work, but I found that the `netgen/configure` is just a wrapper.
And the wrapper has the `CFLAGS` variable hard coded and therefore not overridable on the configure line! 

Why not go with -O3 and forget the -g? If you need to debug, you can always recompile.
I've mentioned changing the optimization to Tim Edwards.

```
#!/bin/sh
# 
# This script is a workaround to GNU autoconf which can't deal with having
# all of its config scripts in a different directory than the configure
# script itself.

export CFLAGS="$CFLAGS -g"
cd scripts
exec ./configure "$@"
```

antonblanchard:
@d-m-bailey the script adds `-g` to the `$CFLAGS` environment variable. Looks to work:

```
gcc -O2 -g -m64 -fPIC -fPIC   -I. -I..  -DCAD_DIR=\"/build/lib\" -DTCL_DIR=\"/build/lib/netgen/tcl\"  -DUSE_TCL_STUBS -DUSE_TK_STUBS -DPACKAGE_NAME=\"netgen\" -DPACKAGE_TARNAME=\"netgen\" -DPACKAGE_VERSION=\"1.3\" -DPACKAGE_STRING=\"netgen\ 1.3\" -DPACKAGE_BUGREPORT=\"eda-dev@opencircuitdesign.com\" -DPACKAGE_URL=\"\" -DNETGEN_VERSION=\"1.5\" -DNETGEN_REVISION=\"161\" -DSTDC_HEADERS=1 -DHAVE_SYS_TYPES_H=1 -DHAVE_SYS_STAT_H=1 -DHAVE_STDLIB_H=1 -DHAVE_STRING_H=1 -DHAVE_MEMORY_H=1 -DHAVE_STRINGS_H=1 -DHAVE_INTTYPES_H=1 -DHAVE_STDINT_H=1 -DHAVE_UNISTD_H=1 -DSIZEOF_VOID_P=8 -DSIZEOF_UNSIGNED_INT=4 -DSIZEOF_UNSIGNED_LONG=8 -DSIZEOF_UNSIGNED_LONG_LONG=8 -DSTDC_HEADERS=1 -DHAVE_SETENV=1 -DHAVE_PUTENV=1 -DHAVE_DIRENT_H=1 -DHAVE_LIMITS_H=1 -DHAVE_VA_COPY=1 -DHAVE___VA_COPY=1 -DDBUG_OFF=1 -DTCL_NETGEN=1 -Dlinux=1 -DSYSV=1 -DISC=1 -DSHDLIB_EXT=\".so\" -DNDEBUG -DX11 -DXLIB -DNDEBUG -DNETGEN_DATE="\"`date`\""  -c netcmp.c
```

We could bump it to `-O3`. I left `-g` because there's no runtime cost, just extra on disk storage.

I also note that magic is being built without any optimisation. I'm more worried there because the code base looks fragile (still using K&R prototypes, and even then it's not always including the prototype).

PR #148, opened 2021-01-04
Speed up extract_antenna_violators.py

antonblanchard:
extract_antenna_violators.py is doing expensive splits across the
entire file which requires large mallocs and memcpys. It also
reads the entire file in before parsing it.

Replace this with a simpler and faster way of achieving the same
thing. On my test case the time goes from over 12 minutes to 0.62
seconds, about 1200x faster.

### DIFF
--- a/scripts/extract_antenna_violators.py
+++ b/scripts/extract_antenna_violators.py
@@ -28,33 +28,25 @@ args = parser.parse_args()
 report_file_name = args.report
 out_file_name = args.output
 
+pattern = re.compile(r'\s*([\S+]+)\s*\([\S+]+\)\s*[\S+]+')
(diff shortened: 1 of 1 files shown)


antonblanchard:
I checked the output was identical on my test case, but someone should double check I am doing the right thing here.

agorararmard (approved):

agorararmard:
@antonblanchard: This is definitely a cleaner/faster way to achieve the same functionality (an even better way would be to do all this with a simple shell script). I tried out a couple of test cases + reviewed the code and I don't see anything wrong with it (for now). So, I'm going to go ahead and merge this PR. Thanks!

PR #149, opened 2021-01-04
fix missing efabless

mattvenn:
the start command doesn't work (starts building another docker image) if the efabless part is missing from the tag.


agorararmard:
@mattvenn: if you're pulling the dockerhub image or changing the local image name then you need to follow this one instead as the documentation states.
https://github.com/efabless/openlane/tree/master#running-the-pulled-auto-built-docker-image 

There is no `efabless/` prefix in the locally built images.

mattvenn:
yeah I just figured it out. I don't know what I did wrong. sorry!

PR #151, opened 2021-01-05
run_designs.py: Create regression_result directory if it doesn't exist

antonblanchard:


agorararmard (approved):

agorararmard:
@antonblanchard: Although it's assumed that run_designs.py has to be run from inside an openlane root project and as such `./regression_results/` must exist because it has documentation and benchmarks stored inside, it doesn't hurt to recursively check for its existence. So, I'm going to merge this PR. Thanks!

PR #168, opened 2021-01-17
Avoid recursively calling generate_final_summary_report()

antonblanchard:
If generate_final_summary_report() fails, we loop forever. We should
harden it against failures but for now clear GENERATE_FINAL_SUMMARY_REPORT.


agorararmard:
Hi @antonblanchard, thanks for the PR, but I don't think this is the right way to tackle the issue at hand. I implemented a fix in 0f936f92de6f1854c6490c40675a30d5e57d5da6 for the first part of the issue (the recursive calling), and another fix for the other half of the issue here: 8beb1411c3fc8a5c992b6d0d63e7654d3eb91799 which allows for better support for generating reports for incomplete runs. Therefore, I'll close this PR.

PR #170, opened 2021-01-18
Write yosys.sdc file

antonblanchard:
I'm trying to fix a timing issue in my design, and I notice we are passing
an empty sdc file to yosys. This fixes it.


ax3ghazy:
Seems that this happened since https://github.com/efabless/openlane/commit/cb0a62f3b59faf6cec188142fa2874e2c4921f4f. Merged, thanks!

PR #180, opened 2021-01-31
Update open_pdks

antonblanchard:
This fixes diode strategy 2 and 5 and patches the Skywater PDK to fix
the MINENCLOSEDAREA DRC issue.


agorararmard:
Hi @antonblanchard, while we appreciate the PR, this process is getting fully automated. That's why we haven't updated the pdk on develop, to continue working on the automation and have something to test with. Check #181, this will be applied on many of the used tools as well, as we move to PR based development and switch everything to master.

PR #204, opened 2021-02-05
Don't use no_synth cells in wire repair

antonblanchard:
Wire repair is using some strange cells for buffers, eg
sky130_fd_sc_hd__lpflow_lsbuf_lh_hl_isowell_tap_1. Avoid
this by using the trimmed library.

### DIFF
--- a/scripts/openroad/or_wireLengthRepair.tcl
+++ b/scripts/openroad/or_wireLengthRepair.tcl
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-foreach lib $::env(LIB_SYNTH_COMPLETE) {
+foreach lib $::env(LIB_OPT) {
     read_liberty $lib
 }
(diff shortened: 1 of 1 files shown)


antonblanchard:
I guess this broke at some point due to a change to OpenROAD:

```
  if { [info exists keys(-buffer_cell)] } {
    utl::warn RSZ 16 "-buffer_cell is deprecated."
  }
```

agorararmard:
@antonblanchard: I pulled your commit and created another PR at #205 to create the LIB_OPT if it's not defined, and to remove the deprecated flag. The command is now working properly. So, I will close this PR.

I will work on merging that command with https://github.com/efabless/openlane/blob/master/scripts/openroad/or_resizer.tcl as both call the same command. I also think we should rely on the `set_dont_use` command from OpenROAD instead of using the `trim_lib` proc, because with the current method we take the timing information away from the tool, which will result in non-efficient result in most of the cases.

PR #217, opened 2021-02-08
Resizer updates

agorararmard:
- This PR merges the buffer insertion script and proc into the existing repair_design proc and script.
- Enable resizer optimizations by default. 
- A separate PR will be made after this is merged to add the new DRC_EXCLUDE_LIST functionality.


scripts/openroad/or_resizer.tcl
```diff
@@ -30,10 +30,11 @@ set_wire_rc -signal -layer $::env(WIRE_RC_LAYER)
 set_wire_rc -clock  -layer $::env(WIRE_RC_LAYER)
 estimate_parasitics -placement
 
-repair_design
-
-#repair_tie_fanout -separation $tie_separation $tielo_port
-#repair_tie_fanout -separation $tie_separation $tiehi_port
+if { $::env(PL_RESIZER_OVERBUFFER) } {
```
ax3ghazy, line 33: How about we eliminate PL_RESIZER_OVERBUFFER and just use MAX_WIRE_LENGTH for this purpose? Otherwise, we need to check if MAX_WIRE_LENGTH exists.

agorararmard: Sounds good. I applied that change.

ax3ghazy (approved):

PR #238, opened 2021-02-15
read lib blackbox, if ADDER Type is RCA, closes #236

### LINKED ISSUE #236: SYNTH_ADDER_TYPE=RCA also requires SYNTH_READ_BLACKBOX_LIB=1
Commit a421bf6689fd ("Add Ripple carry adder/Full adder mapping to yosys (#228)") adds `SYNTH_ADDER_TYPE`. I tried the `RCA` option but it fails with:

```
ERROR: Module `\sky130_fd_sc_hd__fa_1' referenced in module `\adder' in cell `$add$/openLANE_flow/designs/adder/src/adder.v:20$2.stage[0].FA' is not part of the design.
```

Adding access to the standard cells with `set ::env(SYNTH_READ_BLACKBOX_LIB) 1` fixes it. Since the design itself is not using standard cells this is a bit confusing.

The `FA` doesn't have this issue, I'm guessing because we do that mapping later on.

agorararmard:


ax3ghazy (commented):
Need to consider making this the default. Any downsides in sight?

ax3ghazy (approved):
Approving for now...

PR #251, opened 2021-02-21
Tell resizer to buffer inputs and outputs

antonblanchard:
When investigating a timing issue, I noticed a huge amount of delay on
an input. Using AES as an example:

```
Fanout     Cap    Slew   Delay    Time   Description
-----------------------------------------------------------------------------
                  0.00    0.00    0.00   clock clk (rise edge)
                          0.00    0.00   clock network delay (ideal)
                          5.18    5.18 ^ input external delay
                 19.80   14.14   19.32 ^ reset_n (in)
  2739    8.94                           reset_n (net)
                 19.80    0.00   19.32 ^ _35376_/RESET_B (sky130_fd_sc_hd__dfrtp_4)
                                 19.32   data arrival time
```

Enabling input buffering fixes this, and the design makes timing.

### DIFF
--- a/scripts/openroad/or_resizer.tcl
+++ b/scripts/openroad/or_resizer.tcl
@@ -24,6 +24,9 @@ if {[catch {read_def $::env(CURRENT_DEF)} errmsg]} {
     puts stderr $errmsg
     exit 1
 }
+
+read_sdc -echo $::env(BASE_SDC_FILE)
+
 # Resize
 # estimate wire rc parasitics
 set_wire_rc -signal -layer $::env(WIRE_RC_LAYER)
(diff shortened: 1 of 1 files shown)


antonblanchard:
Should we add a config option for this? If so, do we want separate input and output options?

agorararmard:
@antonblanchard: I applied this PR as a patch into my fork https://github.com/agorararmard/openlane/tree/buffer-ports, I added the configuration controls and separated inputs from outputs. I'll start a regression run over the design test set to see the effects of enabling the different combinations and disabling them. Then, I'll re-open another PR. If the results make sense.

So, for now, I'm closing this PR. Thanks!

antonblanchard:
Thanks @agorararmard!

PR #269, opened 2021-03-02
count_lvs.py returns exit code

mattvenn:


agorararmard:
@mattvenn: The current behavior is that count_lvs.py would write the summary file, which would be later parsed for Total Errors, and exit the flow if any was found. This gives us control over what message to print and at the same time allows us to run count_lvs.py with the same try_catch proc used for all other scripts in which we detect any non-zero exit code as a script failure. Unless we rewrite that try_catch proc to accept some exceptions, such as the exit 1 code from count_lvs.py, I will not accept this PR.

Thanks for the PR anyways!

mattvenn:
understood - thanks for the explanation

PR #296, opened 2021-04-21
Enable rootless docker run for openlane

atorkmabrains:
Hi Team,

To run docker, you need to have root access on the machine that you are running on.

We have updated the make file to let it work in case you are using rootless docker setup.


donn:
Sorry, can you please rebase?

donn:
Stale, locked.

PR #301, opened 2021-04-24
CI Overhaul

donn:
This moves the CI entirely to GitHub Actions.

In addition to the initial porting effort by @agorararmard, the CI has been rearchitected and nontrivial scripts were rewritten in Python for maintainability. Full documentation of the rearchitected CI is to be done, but highlights include:
- Repository-independence: The Github Actions CI works whether you're efabless/openlane or a fork, and tests treat all PRs equally
- Docker Hub is no longer used for intermediate data storage, rather, GA's artifacts are used for that. This allows external PRs to be tested without any workarounds.
- Tag updates no longer happen in pull requests. The latest tag is also pushed to docker as "current", so "current" can be pulled from Docker Hub. Makefiles have been updated accordingly.
- PDK and tool branches only tested after the pull request is created (to avoid double-testing and creating useless Docker images).
- OpenCircuitDesign.com's git repositories have been replaced with their GitHub mirrors to improve CI stability (sorry @RTimothyEdwards )

--
Notes:
- A typical push/PR test takes 30 minutes. A typical scheduled, daily test takes 4 hours.
- The GitHub account https://github.com/openlane-bot must be added as core developer prior to the merging of this pull request. This bot's token is used as secrets.MY_TOKEN and naturally if it has no write access to efabless/openlane then the CI will fail. (cc. @mkkassem )
- Please **squash and merge**. There are far too many commits in this repository.

PR #302, opened 2021-04-24
Repo skywater-pdk has replaced branch master with main

russellfriesenhahn:


donn:
Will be addressed as part of #301.

PR #322, opened 2021-05-04
Update synth.tcl

mattvenn:
this creates a useful dot file to examine the techmap process. It's done before too much optimisation so the results are understandable for small designs.


donn:
Looks good. Thanks Matt.

PR #341, opened 2021-05-11
Exclude the cells with drc erros from CTS liberty file

Manarabdelaty:
After running CTS, I got `sky130_fd_sc_hd__lpflow_lsbuf_lh_hl_isowell_tap_1` cell inserted in the def file which is part of the drc exclude list. This issue appears to happen on large designs (like caravel `mgmt_core`).


donn:
Squashed because of some issues, but looked fine otherwise.

PR #353, opened 2021-05-17
cdl file adapting to both ef_formatted paths and regular ones

ahmednofal:


donn:
Master is frozen. Sorry.

PR #384, opened 2021-05-25
Fix the URL for the openlane releases

leoheck:
This link was broken.


donn:
Added to #435- thanks

PR #392, opened 2021-05-26
Customize docker options when using 'podman'

mballance:
Centos8 installs 'podman' instead of Docker. Compatibility wrappers are provided such that docker entry-points can still be used. One big difference is that podman doesn't accept the user specification (-u user:group) that Docker does. I've been using podman successfully with Openlane just by not passing the -u... option.
This PR omits passing the -u option to docker if podman is actually being used.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,6 +35,13 @@ ifeq (, $(strip $(NPROC)))
 
 endif
 
+# Podman (Centos8) doesn't like the -u switches
+# Only add if we're not using podman in emulation
+ifeq (,$(shell which podman))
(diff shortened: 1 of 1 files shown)


donn:
Have a small concern: you seem to be checking for the existence of podman period rather than checking if the installation is aliasing docker. Unless docker and podman cannot be installed on the same system?

mballance:
Good point, @donn. While I think it's extremely uncommon to have both installed, I can't say it's impossible. I'll update to actually invoke the 'docker' command and check whether it reports actually being podman.

donn:
Looks good, thanks @mballance!

PR #412, opened 2021-05-31
Exclude DRC list from CTS liberty file + Apply sed command on the SYNTH_NETLIST

Manarabdelaty:


donn:
Can you also apply these fixes to `develop`?

Manarabdelaty:
@donn They are already part of develop.

jeffdi (approved):

PR #422, opened 2021-06-03
Add include paths when processing top-level and blackbox verilog files

mballance:
Some of the IPs in my design define ports using macros declared inside include files. For example:

```
`include "wishbone_tag_macros.svh"
`include "wishbone_amo_defines.svh"
  
module fwrisc_rv32imca_wb #(
		parameter VENDORID 	= 0,
		parameter ARCHID 	= 0,
		parameter IMPID 	= 0
		) (
		input				clock,
		input				reset,
		input[31:0]			hartid,
		`WB_INITIATOR_TAG_PORT( , 32, 32, 1, 1, 4),
		input				irq
		);
```
This PR passes include paths to the processing of black-box and top-level verilog files, allowing these included files to be found.

Signed-off-by: Matthew Ballance <matt.ballance@gmail.com>

### DIFF
--- a/scripts/synth.tcl
+++ b/scripts/synth.tcl
@@ -30,6 +30,14 @@ if { [info exists ::env(SYNTH_DEFINES) ] } {
 	}
 }
 
+set vIdirsArgs ""
+if {[info exist ::env(VERILOG_INCLUDE_DIRS)]} {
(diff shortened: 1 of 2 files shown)


donn:
Looks good, thanks

Issue #424, opened 2021-06-05
Add facility to package failed runs


maliberty:
In OpenROAD-flow-scripts the 'make issue' facility makes it easy to package up all necessary files to reproduce a problem in order to reproduce an issue.  OpenLane needs a similar capability.  See the on-going discussion in https://github.com/The-OpenROAD-Project/OpenROAD/issues/703.

donn:
I'll be honest, I have no idea what on earth is going on here: https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/blob/b549b82ff7a3c004a0117ebea2ef970f2d4c7207/flow/util/utils.mk#L57

The (unofficial) ruby script I made simply tries to extract the variables and files, then replace them with resolved values. Does this do something similar?

maliberty:
It produces a tgz with runme script that (1) sources a script that captures all the envars set by make and (2) invokes openroad on the proper tcl file to reproduce the step that is failing.  That tgz is a complete capture of the problem and is trivial for a developer to unpack and run.

donn:
Huh. I like the idea of the runner script being the one to set the environment variables. I'll look into it.

PR #428, opened 2021-06-07
or_issue.py (RE: The-OpenROAD-Project/openlane#424)

donn:
This adds the capability to package OpenROAD scripts into a self-contained fileset that can be used to reproduce an error in a vaccuum.

A minor fix was applied to `or_replace.tcl` to prevent it from copying the entire scripts folder.

### DIFF
--- a/or_issue.py
+++ b/or_issue.py
@@ -1,0 +1,210 @@
+#!/usr/bin/env python3
+
+# Copyright 2021 Efabless Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import os
+import re
+import sys
+import shutil
+import pathlib
+import argparse
+from collections import deque
+from os.path import join, abspath, dirname, basename, isdir, relpath
+
+openlane_path = abspath(dirname(__file__))
+
+parser = argparse.ArgumentParser(description="""
+This script creates a reproducible, self-contained package of files to demonstrate
+OpenROAD behavior in a vaccum, suitable for filing issues.
+
+The result is a tarred, xzipped and gzipped tarball for minimal size while
+maintaining GitHub uploadability.
+
+Requires UNIX-like operating system. Final package printed to /dev/stdout. Don't forget to chomp/rstrip.
+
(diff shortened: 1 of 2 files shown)


or_issue.py
```diff
@@ -0,0 +1,210 @@
...
+
+def read_env(config_path: str, from_path: str, input_env={}) -> dict:
+    rx = r"\s*set\s*::env\((.+?)\)\s*(.+)"
+    env = input_env.copy()
+    string_data = ""
+    try:
+        string_data = open(config_path).read()
+    except FileNotFoundError:
+        print(f"❌ File {config_path} not found. The {from_path} path may have been specified incorrectly.", file=sys.stderr)
+        exit(66)
```
maliberty, line 78: What does 66 denote?

or_issue.py
```diff
@@ -0,0 +1,210 @@
...
+
+# Phase 5: Create Run File
+run_cmd = join(destination_folder, "run")
+with open(run_cmd, "w") as f:
+    env_list = "\\\n    ".join([f"-e {key}='{value}'" for key, value in final_env_pairs])
+    f.write(f"""
+#!/bin/sh
+dir=$(cd -P -- "$(dirname -- "$0")" && pwd -P)
+cd $dir;
+docker run --rm\\
```
maliberty, line 196: It would be helpful to have a non-docker option

or_issue.py
```diff
@@ -0,0 +1,210 @@
...
+    {env_list}\\
+    {os.getenv("IMAGE_NAME") or "efabless/openlane:current"} openroad \\$::env\\(OR_SCRIPT\\)
+    """)
+os.chmod()
+
+# Phase 6: Tarball and output
+last_output = destination_folder
+if tarball:
+    last_output = destination_folder + ".tar.gz"
+    os.system(f"tar -cv {destination_folder} | xz | gzip > {last_output}")
```
maliberty, line 207: Why apply xz & gzip?  Also xz is quite slow so it would be good to make this an option.

maliberty (commented):
I don't have a lot of experience with the OL structure but this look reasonable.

or_issue.py
maliberty: Also the name should be .tar.xz.gz

or_issue.py
donn: EX_NODATA, a [BSD exit code](https://www.freebsd.org/cgi/man.cgi?query=sysexits&apropos=0&sektion=0&manpath=FreeBSD+4.3-RELEASE&format=html). It's probably the most ignored standard in existence at this point but I just support it where I can.

Manarabdelaty (approved):

or_issue.py
maliberty: Use os.EX_NOINPUT in that case (there is no EX_NODATA afaik)

donn: Yeah, sorry, that was a pre-coffee message. I didn't know python enumerated sysexits as well, that's kinda cool

or_issue.py
maliberty: I don't see any issue with uploading a .xz.  See https://github.com/maliberty/xz_test

maliberty: Sorry I missed that the issue is attaching to an issue and there the list is oddly limited.

PR #433, opened 2021-06-10
Migrate to a more current version of OpenROAD

donn:
This PR is a collaborative effort between the OL and OR teams to bring OpenLane up to speed with the current version of OpenROAD.

Specifically, OpenROAD commit 4d4d7205fd0292dbf3fae55fad9109b3f0bd5786.

This commit also removes all standalone utilities that were once integrated into OpenROAD.

I advise that this issue be **Squashed and Merged**.


osamahammad21:
Hi @donn 
This clearly has some good insights towards updating openroad version in openlane. From looking at the changes, here are some of the comments I have that we need to address:
- In this version, OpenDB_python is removed but openlane clearly use it.
- We should remove the patches since they are not used.
- PL_DIAMOND_SEARCH_HEIGHT should be removed from configuration

There are other updates that need to be done to other or_scripts(including cts, antenna, routing, etc...). Here is a branch that has some of these updates [osamahammad21:update-or](https://github.com/osamahammad21/openlane/tree/update-or). I've had it on my local for a while until I address some issues in OpenROAD.
Also I had some questions I was planning to ask openlane folks about:
- Are the changes made by patches still needed or all fixed? specially pdn-export-subst and rails
   - Why do you use patches not a fork? 
- Tile size and unidirectional routing options are not supported anymore by OpenROAD GRT, can we remove them?
- Did you verify the spef extractor correctness? It seemed to fail when the spacing changed in the def file(still valid def) which indicates that it might be prone to errors.

donn:
Thanks @osamahammad21: Could you open a Draft PR (I won't merge it.)? Getting a diff would go a long way here.

> In this version, OpenDB_python is removed but openlane clearly use[s] it.
It works fine though.
> We should remove the patches since they are not used.
> Are the changes made by patches still needed or all fixed? specially pdn-export-subst and rails + Why do you use patches not a fork?
I have to make sure that everything works without the patches first. This is still a WIP. The reason I believe was so we don't have to maintain a fork, we can just grab OpenROAD master, patch it and go to town. Forks are annoying.
> PL_DIAMOND_SEARCH_HEIGHT should be removed from configuration.
I'm keeping some documentation that we removed it and it no longer has an effect. I'll see where else is it being used.
> Tile size and unidirectional routing options are not supported anymore by OpenROAD GRT, can we remove them?
Will do. I already disabled them in the scripts I thing.
> Did you verify the spef extractor correctness? It seemed to fail when the spacing changed in the def file(still valid def) which indicates that it might be prone to errors.
No :)

osamahammad21:
@donn

> In this version, OpenDB_python is removed but openlane clearly use[s] it.

>> It works fine though.

Actually it doesn't. If you check openlane CI, it fails because It can't find the [OpenDB_python](https://github.com/The-OpenROAD-Project/OpenLane/pull/433/checks?check_run_id=2789104306#step:11:1329) dir. It could have worked fine on your local on make merge because the dir is already there.
>Did you verify the spef extractor correctness? It seemed to fail when the spacing changed in the def file(still valid def) which indicates that it might be prone to errors.

>>No :)

Why not use OpenROAD's command.

donn:
> Actually it doesn't. If you check openlane CI, it fails because It can't find the OpenDB_python dir. It could have worked fine on your local on make merge because the dir is already there.

Not talking about a specific folder: `openroad -python` then `import opendb` works fine. This is the way I was told to handle Opendbpy in the future by @maliberty and should this feature be unintentional in any form it would have to be surfaced.

> Why not use OpenROAD's command

Only recently started maintaining this codebase. There are going to be an odd number of intricacies, including but not limited to workarounds for behavior that may have been different or missing on the version of OpenROAD we're using. I'm not opposed to changing to the the OpenROAD command.

--

Side note, the PDN patches will have to be upstreamed in some form. Running into too many issues without them.

maliberty:
@donn if patches need to be upstreamed you should submit a PR

donn:
Yeah, I will first thing tomorrow I think.

donn:
Oh for the love of

donn:
This is mostly fine now. Last issue is #482, which we can deal with at our own pace.

PR #500, opened 2021-07-26
Enable rootless docker for more secure usage of the Openlane

atorkmabrains:
Enable rootless docker for more secure usage of the Openlane


donn:
Hmm, this... is breaking the PDK installation in some manner. I've run it multiple times.

atorkmabrains:
Hi @donn 
Are you testing it from CI here? It works on Ubuntu 20.04. I'm not sure why it's failing the CI here.

Regards,
Amro

donn:
Guessing it's a permissions issue.

Sorry, if you want this merged you'd have to fix the CI as well.

donn:
I'm afraid we'll have to reject this PR.

As it stands, there are too many configurations that a one-size-fit all does not appear to be the best solutions, especially that the vast majority of our users and customers use a regular Docker setup for which changing the permission configuration to `-u 0` would result in myriad permissions issues, despite the superior security implications. Had this not been caught by the CI, I would have been dealing with a nightmare.

I personally use a daemonless, podman-based setup on my own machine.

If you have a mechanism by which you can differentiate a rootless Docker setup from an orthodox one and alter the permission configuration flags accordingly, please add it to https://github.com/The-OpenROAD-Project/OpenLane/blob/master/get_docker_config.py, which similarly differentiates Docker and Podman (Podman interpreting permission configurations very differently from Docker), we would be happy to merge it.

Issue #493, opened 2021-07-21
save_state crash when saving environment variables that contain double quotes


getziadz:
**Description**
OpenLane crashes when running the example design.
The source of the crash is the save_state function which crashes when it encounters environment variables with double quotes

**To Reproduce**
Define an environment variable that contains double quotes, such as:
`function urg() { command urg -full64 "$@"; }; export -f urg`

run
`make test`

You will observe the following:
```
[INFO]: Saving Runtime Environment
extra characters after close-quote
    while executing
"set ::env(BASH_FUNC_urg()) "() {  command urg -full64 "$@"
}""
    ("uplevel" body line 1)
    invoked from within
"uplevel #0 ${cmd}"
    (procedure "set_log" line 3)
    invoked from within
"set_log ::env($index) $::env($index) $::env(GLB_CFG_FILE) 1"
    (procedure "save_state" line 8)
    invoked from within
"save_state"
    (procedure "run_non_interactive_mode" line 69)
    invoked from within
"run_non_interactive_mode {*}$argv"
    invoked from within
"if { [info exists flags_map(-interactive)] || [info exists flags_map(-it)] } {
        puts_info "Running interactively"
        if { [info exists arg_values(-file)..."
    (file "./flow.tcl" line 346)
```

**Expected behavior**
OpenLane crashes

**Desktop (please complete the following information):**
 - OS: Springdale Linux release 7.9 (Verona)
 - Openlane Version:
   - v0.22-3-g7b3af66
   - master branch ( 02c16938aea3770c1d2e03fc8beed763fa83f30f )

**Additional context**
A simple solution is to escape the double quotes. See https://github.com/getziadz/OpenLane/commit/ec2d52f5df8d7fe30f04323931b4be71f235fc8e

donn:
Thank you so much @getziadz- I'll merge the fix within the next few days.

getziadz:
No problem. Should I create a pull request?

donn:
That would greatly accelerate things, yes.

getziadz:
Just created the pull request.

donn:
Thank you so much.

ahmednofal:
I believe this to be still persistent in version 2021.09.19_20.25.16  
This is the traceback from the crash.
```
invalid command name ""/home/nofal/design/project_using_sky130A/verilog/source/spm.v""
while executing
"\"/home/nofal/design/project_using_sky130A/verilog/source/spm.v\""
[...]

PR #502, opened 2021-07-26
Patch for save_state function: crash due to environment variables wit…

getziadz:
…h double quotes

'save_state' function crashes when trying to save environment variables that contain double quotes

Example definition (.bashrc) that causes a crash:

function vcs() { command vcs -full64 "$@"; }; export -f vcs

Fix for issue: https://github.com/The-OpenROAD-Project/OpenLane/issues/493


donn (approved):
Looks good!

PR #510, opened 2021-07-28
Versioning Improvements

### LINKED ISSUE #469: Disable OpenPhySyn
OpenPhySyn is no longer maintained or developed.

### LINKED ISSUE #490: make openlane always pulls efabless/openlane:current irregardless of the openlane version
This is creating an issue where an older version of openlane is used and `make openlane` is pulling a docker image that doesn't correspond to that version.

donn:
Getting the version for OpenLane now tries to use Git, and if it fails, prompts the user for a version. Fixes #490.

There is now one central versioning manifest for the tool the docker build, `make pdk` and installer all get data from.

Additionally, OpenPhySyn has been removed. Fixes #469.

New versions of OpenLane will be nightly and timestamped in a YYYY.MM.DD_HH:mm:SS format.


donn:
As the issue spans many commits, it is recommended to Squash and Merge.

maliberty (approved):
I'm not much of an OL expert but it looks ok to me.  FYI, OR keeps its dependencies in https://github.com/The-OpenROAD-Project/OpenROAD/blob/master/etc/DependencyInstaller.sh

PR #519, opened 2021-08-02
OR Hotfix

donn:
As the internal APIs of OR aren't stable, some scripts like or_groute, which rely heavily on the internal APIs, will break.

Until this reliance on the internal APIs is dealt with, I'll individually fix any issues that may arise.


donn:
This is gonna need more massaging than I anticipated.

PR #520, opened 2021-08-02
Update OpenROAD and Klayout Scripts

donn:
* Remove some OR commands that are no longer available
* Leverage Klayout Batch Mode instead of xvfb-run
* Update READMEs to show removed options as **removed** explicitly: I'm sure users would not appreciate options just disappearing with no paper trail


donn:
Too many issues. Aborting this for now.

PR #531, opened 2021-08-10
Added timing optimizations after global routing

marwaneltoukhy:


donn:
Appears to have broken macro placement. Can you investigate?

donn (changes requested):
There are multiple configuration values added that remain undocumented.

donn (approved):
Changes look ok now, will merge if CI passes

PR #539, opened 2021-08-16
Add OpenRCX support

Manarabdelaty:
- OpenRCX is used by default as the primary method for spef extraction
- DEF2SPEF is still supported and can be selected by setting `SPEF_EXTRACTOR` to `def2spef`
- Waiting on this PR (https://github.com/RTimothyEdwards/open_pdks/pull/145) to be merged to open_pdks


donn:
I’m not sure if it entered an infinite loop of some kind but the fastest test set has been running for 3 hours now

Manarabdelaty:
That sounds likely, I am isolating whether it is the pdk update or something with openrcx. I am thinking it is the pdk update because I have been running the regressions with openrcx with no issue

donn (approved):
Looks A-OK now

Issue #535, opened 2021-08-13
Provide support for Static IR Drop analysis in OpenLane similar to OpenROAD flow scripts


vijayank88:
Hi OL team,
@maliberty  @donn @Indira
Issue is:
I want to generate static IR drop for my design. In OpenRoad flow/scripts/final_report.tcl has analyze_power_grid option to get IR drop details.

But if I try to get in OpenLane flow its throw an error as command not found.

PDNSim GitHub pages says its archived and part OpenRoad flow.

Enhancement Looking:
OpenRoad scripts integrated with OpenLane flow(Named as OpenRoad app). Also integrate the PDNSim feature in OpenLane flow scripts to generate IR Drop Details.

Thanks in advance...

vijayank88:
[6-pdn.log](https://github.com/The-OpenROAD-Project/OpenLane/files/6986430/6-pdn.log)

Command used in OpenRoad script: analyze_power_grid -net VDD will generate IR drop details for this power domain
If I use analyze_power_grid -net VDD in OpenLane its stop with error as met3 resistance value not found in DB.
How to reproduce the issue: 
Run spm design in OpenLane flow with below modification in or_pdn.tcl:

analyze_power_grid -net $::env(VDD_NET)

add above line in scripts/openroad/or_pdn.tcl after check_power_grid line. 

Solution Expected:
Generate IR Drop for power domain network.

vijayank88:
With openlane tag 2021.09.09_03.00.48 IR drop calculation updated. we can see IR drop in 7-pdn.log.

PR #549, opened 2021-08-19
Allow IR Drop Calculation

donn:
Addressing #535.

--
 
Adds ability to optionally calculate IR drop (user must provide file with rc values).
* A discussion should be held in the near future on if RC values should be calculated as part of open_pdks.

* Change PDN_CFG so it's included in the generated config.tcl file (for or_issue.py)

* Fix The-OpenROAD-Project#495

The **APU** design gives an example of how to enable IR drop calculation.

### DIFF
--- a/designs/APU/rc.tcl
+++ b/designs/APU/rc.tcl
@@ -1,0 +1,48 @@
+# From OpenROAD Flow Scripts
+
+# BSD 3-Clause License
+
+# Copyright (c) 2018, The Regents of the University of California
+# All rights reserved.
+
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+
+# * Redistributions of source code must retain the above copyright notice, this
+#   list of conditions and the following disclaimer.
+
+# * Redistributions in binary form must reproduce the above copyright notice,
+#   this list of conditions and the following disclaimer in the documentation
+#   and/or other materials provided with the distribution.
+
+# * Neither the name of the copyright holder nor the names of its
+#   contributors may be used to endorse or promote products derived from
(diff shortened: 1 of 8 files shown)


designs/APU/rc.tcl
```diff
@@ -0,0 +1,48 @@
...
+# sync from openroad repo test/sky130hd/sky130hd.rc
+set_layer_rc -layer li1 -capacitance 1.499e-04 -resistance 7.176e-02
+set_layer_rc -layer met1 -capacitance 1.449e-04 -resistance 8.929e-04
+set_layer_rc -layer met2 -capacitance 1.331e-04 -resistance 8.929e-04
+set_layer_rc -layer met3 -capacitance 1.464e-04 -resistance 1.567e-04
+set_layer_rc -layer met4 -capacitance 1.297e-04 -resistance 1.567e-04
+set_layer_rc -layer met5 -capacitance 1.501e-04 -resistance 1.781e-05
+# end sync
+
+set_layer_rc -via mcon -resistance 9.249146E-3
```
Manarabdelaty, line 41: @donn Do you think we can move this information to `open_pdks` since it is pdk specific ? The vias resistance information is also needed in openrcx and they are part of this PR (https://github.com/RTimothyEdwards/open_pdks/pull/145/files#diff-5fdc6252f3c9219615713f60ac9aa89db9010959caba3daed54187c0d80c4d45R127)

donn: This would actually be better, so yes. Shall we discuss this further in our meeting today?

Manarabdelaty: OK

Issue #501, opened 2021-07-26
set_propagate_clock [all_clocks] not working


vijayank88:
**Describe the bug**
in OpenLane flow scripts/openroad/or_resizer_timing.tcl  below lines available. 

```
# CTS and detailed placement move instances, so update parastic estimates.
estimate_parasitics -placement
set_propagated_clock [all_clocks]
repair_timing
```

The respective log is 15-fastroute.log don't have propagated delay, still running with ideal delay. Log shown below:

```
Startpoint: _898_ (rising edge-triggered flip-flop clocked by clk)
Endpoint: _884_ (rising edge-triggered flip-flop clocked by clk)
Path Group: clk
Path Type: max

Fanout     Cap    Slew   Delay    Time   Description
-----------------------------------------------------------------------------
                  0.00    0.00    0.00   clock clk (rise edge)
                          0.00    0.00   **clock network delay (ideal)**
                  0.00    0.00    0.00 ^ _898_/CLK (sky130_fd_sc_hd__dfxtp_1)
                  0.06    0.84    0.84 v _898_/Q (sky130_fd_sc_hd__dfxtp_1)
     2    0.00                           uart_rx_inst.prescale_reg[1] (net)
                  0.06    0.00    0.84 v _485_/A (sky130_fd_sc_hd__or2_1)
                  0.18    0.59    1.43 v _485_/X (sky130_fd_sc_hd__or2_1)
     4    0.01                           _328_ (net)
                  0.18    0.00    1.43 v _486_/C (sky130_fd_sc_hd__or3_1)
                  0.22    0.87    2.31 v _486_/X (sky130_fd_sc_hd__or3_1)
     4    0.01                           _329_ (net)
                  0.22    0.00    2.31 v _487_/D (sky130_fd_sc_hd__or4_1)
                  0.26    1.12    3.43 v _487_/X (sky130_fd_sc_hd__or4_1)
     3    0.01                           _330_ (net)
```
**To Reproduce**

Take any available design in OpenLane flow and check design/<name>/runs/logs/routing/fastroute.log

**Expected behavior**

Once we add set_propagated_clock [all_clocks] this clock dealy should be used in all next stages. If we included in cts, routing stage should be used same propagated clocks. Example log shown below:
```
Startpoint: _897_ (rising edge-triggered flip-flop clocked by clk)
Endpoint: _874_ (rising edge-triggered flip-flop clocked by clk)
Path Group: clk
Path Type: max

[...]

PR #562, opened 2021-08-28
fix for GitHub Issue #501 on adding propagation clock

vijayank88:
I have updated or scripts to write sdc file at CTS stage to use the updated sdc file for timing check post CTS with propagated delay. This is part of #501 fix. I checked #make fastest_test_set flow completed without fail.


maliberty:
It seems ok to me but I'll leave it to @donn to approve as I don't normally approve OL changes.

donn (approved):
it looks fine

PR #566, opened 2021-08-30
fetch the correct tritonRoute def file for diearea report calculation

kareefardi:
Previously, `openlane/.../<run>/results/routing/${designName}.def` (doesn't exist) was selected as tritonRoute def file to calculate the diea area. Now, it is changed to fetch the correct variables similar to the rest of variables in `report.sh` .


donn (approved):
Changes look ok, will merge if CI passes

PR #575, opened 2021-09-01
Fixing some issues in report generation

kareefardi:
1. Save state before generating the report. Generating the report reads a run's `config.tcl` which is changed after saving the state, so we should save the state first then generate the report.
2. In `report.sh`, fetching some values happens before generating the values themselve. This adjusts the sequence of commands properly. 
3. Report `E404` instead of `-1` in values that fail to get captured as some of these values can be `-1` themselves.
4. Some indenting in `report.sh`

### DIFF
--- a/flow.tcl
+++ b/flow.tcl
@@ -123,7 +123,7 @@ proc run_antenna_check_step {{ antenna_check_enabled 1 }} {
 	}
 }
 
-	
+
 
 proc run_non_interactive_mode {args} {
 	set options {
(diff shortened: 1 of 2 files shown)


donn (approved):
more alarmed that you went through all these damn changes without just rewriting that file in python

kareefardi:
> more alarmed that you went through all these damn changes without just rewriting that file in python

They were mostly cut and paste

PR #583, opened 2021-09-02
Fix PIP issue with rtd

donn:
This fixes an issue with pip where support for file: was abruptly removed.


kareefardi:
Just a quesiton. There is no potential issue here using `. ` for relative pathing?

donn:
I don't believe so,  no.

kareefardi (approved):

PR #588, opened 2021-09-06
Rework Multithreading Variables

### LINKED ISSUE #509: ROUTING_CORES not working
In my design/.../config.tcl I set:
set ::env(ROUTING_CORES) 40

However when the designs runs in TritonRoute I don't see 40 threads working.

**To Reproduce**
Run any design with this modification.

I used OL 3c0f2fd0e5daf1ff1047fa802b7cc53bb8d4f7a7 as my build.

donn:
~ ROUTING_CORES now overridable by environment (set_if_unset)
- Deprecated TR parameter file removed

Fixes #509


configuration/routing.tcl
```diff
@@ -13,7 +13,9 @@
 # limitations under the License.
 
 # Routing defaults
-set ::env(ROUTING_CORES) 4
+if {! [info exists ::env(ROUTING_CORES)] } {
+    set ::env(ROUTING_CORES) 2
```
maliberty, line 17: Why is the default changed from 4 to 2?  The documentation doesn't reflect that change.

scripts/openroad/or_droute.tcl
```diff
@@ -24,7 +24,14 @@ if {[catch {read_def $::env(CURRENT_DEF)} errmsg]} {
...
 set_thread_count $::env(ROUTING_CORES)
 
-detailed_route -param $::env(tritonRoute_tmp_file_tag).param
+detailed_route\
+    -guide $::env(CURRENT_GUIDE)\
+    -output_guide $::env(TRITONROUTE_FILE_PREFIX).guide\
+    -output_maze $::env(TRITONROUTE_FILE_PREFIX)_maze.log\
+    -output_drc $::env(TRITONROUTE_RPT_PREFIX).drc\
+    -droute_end_iter $::env(ROUTING_OPT_ITERS)\
+    -or_seed 42\
```
maliberty, line 33: Why are you using or_seed?

configuration/routing.tcl
donn: No reason. I’ve dealt with my fair share of dual cores.

scripts/openroad/or_droute.tcl
donn: Fixing the seed gives the test a lot more replicability. I’m not sure if you fix a seed by default and I don’t feel like dedicating time testing that.

configuration/routing.tcl
maliberty: Unless you plan to change the docs I suggest you leave it at 4.  Even 4 seems a bit low for modern computers.

scripts/openroad/or_droute.tcl
maliberty: The seed is fixed - this is for trying to see if alternative values give a better result.

configuration/routing.tcl
donn: Docs have been changed. When running multiple designs in parallel on cloud environments, two cores per design is quite reasonable… people with more resources can simply override the relevant environment variable.

scripts/openroad/or_droute.tcl
donn: Eh, well, this particular seed has come up with our current benchmarks so unless you’re planning to deprecate this option I will not rock the boat unnecessarily.

Manarabdelaty (approved):

PR #592, opened 2021-09-07
Fix for save_state function

getziadz:
The escaped_env_var is not reused correctly so that all escaped characters are correctly escaped


kareefardi (approved):
Yes that seems correct. Thank you

PR #597, opened 2021-09-08
revert E404 to -1 temporarily

kareefardi:
There is a lot of dependencies on `-1`. Will keep it like that for now, until I handle it dependencies cleanly


donn:
Moot, #608

PR #605, opened 2021-09-13
Added IR Drop Calculation To PDN Step

donn:
This adds the ability to optionally calculate the IR drop as requested in #535.

A small report as added to the end of `7-pdn.log`:

```
########## IR report #################
Worstcase voltage: 1.80e+00 V
Average IR drop  : 2.63e-09 V
Worstcase IR drop: 8.32e-08 V
######################################
```

---

Some notes for @Manarabdelaty before approval or merge:

- WIRE_RC_LAYER appears to have been added to the PDK at some point, so I removed it from the OpenLane config
- I moved DATA/CLOCK_WIRE_RC_LAYER to be "independent" as they are used in more than one step now.
    - Should these also be moved to the PDK?
- FP_PDN_IRDROP was added, but it's turned off by default, should it be turned on?

### DIFF
--- a/configuration/README.md
+++ b/configuration/README.md
@@ -16,6 +16,12 @@ This page describes configuration variables and their default values.
 
 These variables are optional that can be specified in the design configuration file.
 
+### Multiple
+| Variable      | Description                                                   |
+|---------------|---------------------------------------------------------------|
+| `DATA_WIRE_RC_LAYER` | The metal layer used for estimating the resistance/capacitence values for data nets. <br> (Default: `met2`)|
+| `CLOCK_WIRE_RC_LAYER` | The metal layer used for estimating the resistance/capacitence values for clock nets. <br> (Default: `met5`)|
+
 ### Synthesis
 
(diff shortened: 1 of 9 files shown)


Manarabdelaty:
@donn Yes, I think it makes more sense to move the DATA/CLOCK_WIRE_RC_LAYER  to the PDK. Also, I agree with having the IR drop calculation enabled by default since it is a standard way of evaluating the PDN.

donn:
Feedback from Shalan (in order of priority)

* Report should be isolated: Currently it's lumped in with `7-pdn.log`
* The output should be processed and the user should be warned if needed (merge later?)
     * Beyond 5% should be reported, value is a simple OpenLane parameter
* STA should be done using nearest corner to worst case as well

We'll probably merge the first then work on the rest.

donn:
PDK has been updated to latest. Manar, I think this is ready.

Manarabdelaty (approved):

PR #609, opened 2021-09-16
Fixes for the multi power domain generation, custom IO placement, and diode insertion strategy 4

Manarabdelaty:
- Drawing multiple power domains wasn't working properly because `write_powered_def.py` assumed that the default vdd and gnd nets always exist at index 0. This isn't always the case because the nets are ordered differently in the def file. This PR fixes the issue by enforcing the index of the default power and ground net (the default power domain is assumed to be the first net specified by `::env(VDD_NETS)` and `::env(GND_NETS)`). 
- Added two new variables to give more control on the macros connection to the top level power grid.  `FP_PDN_ENABLE_MACROS_GRID` will either enable or disable connecting internal macros to power. 
`FP_PDN_MACRO_HOOKS` gives the ability to specify explicit power connections for internal macros. It is useful when the design has multiple power domains. For example, 
set ::env(FP_PDN_MACRO_HOOKS) "\
	mprj vccd1 vssd1"
will connect macro with instance name `mprj` to the power domain of `vccd1` and `vssd1`. This depends on https://github.com/RTimothyEdwards/open_pdks/pull/161 
- Added @20Mhz fix for the custom IO placement 
- Updated the `place_diodes.py` to remove the dependency on the `new_int` API which is no longer available in openroad.


donn (approved):
Everything looks okay, will merge if/when CI clears

PR #610, opened 2021-09-16
Upload logs as GHA Artifacts

donn:
Unai confirmed that from his experience, GHA Artifact storage for public repositories is unlimited.

Given the limits of GCP (can't upload artifacts from PRs,) we're just going to move to this.

### DIFF
--- a/.github/scripts/generate_tag.py
+++ b/.github/scripts/generate_tag.py
@@ -35,13 +35,12 @@ for tag in gh.openlane.tags:
 commit_count = int(subprocess.check_output(["git", "rev-list", "--count", "%s..%s" % (latest_tag_commit, "HEAD")]))
 
 if commit_count == 0:
-    print("No new commits.")
-    gh.export_env("NEW_TAG", "NO_NEW_TAG")
-    exit(0)
+    print("No new commits. A tag will not be created.")
+else:
+    now = datetime.datetime.now()
 
-now = datetime.datetime.now()
+    new_tag = now.strftime("%Y.%m.%d_%H.%M.%S")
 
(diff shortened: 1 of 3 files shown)


mithro:
@donn -- I suggest that we upload to **both** GHA Artifact storage and GCP.

I believe we should be able to figure out a way to upload GCP artifacts from PRs -- @QuantamHD / @kgugala any thoughts? Can the GitHub Action Self Hosted runners do something extra to make this work?

donn:
Not a bad idea. For PRs, as soon as we get around to the GCP move then yeah it would work with some permission work but as long as we’re on GHA, there’s no way to do that cleanly. As a temporary solution I can simply keep GCP log uploads for the daily runs at least.

QuantamHD:
I mean we can certainly upload things to github buckets and make them publicly available. Is there a security or access concern @donn? What is the customer user journey of someone making use of the logs?

QuantamHD:
We can certainly create a service account that has permissions to upload to a specific bucket.

mithro:
@quantamhd - I think the issue is in a pull request someone could exfiltrate the service key and start uploading random stuff?

donn:
@QuantamHD We already have a system like that in place. Problem is, the GCP token we’re using is a GitHub Secret. Secrets are not exposed to pull requests because a bad actor can just POST the token somewhere.

QuantamHD:
Cool! That was always a question I had a about the security model of GHA. You could setup some watch on GHA artifacts and mirror them in GCP, but that just seems like extra work. Once we setup the dedicated self hosted runners I imagine we can provide our own storage backend in GCP without relying on github secrets.

shalan (approved):

PR #624, opened 2021-09-24
LIB_SYNTH has no default value in openlane

ahmednofal:


donn:
The default is set by the PDK in open_pdks, not OpenLane. However, there is a case to be made that it could be "PDK-dependent" in the documentation.

PR #633, opened 2021-10-02
Add chip_integration.md variables to configuration/README.md (they were missing)

ahmednofal:
Some variables found in chip_integration.md were not documented. They are now.


donn:
I pulled the diff and will integrate with an upcoming patch. thx

PR #650, opened 2021-10-17
Fix minor logic issue

donn:


vvbandeira:
@donn approved as requested

PR #654, opened 2021-10-20
Simple CTS by default

donn:
Note: These still fail because the clock periods need to be adjusted.


donn:
Motivation was discussed at length during the sync meeting.

maliberty:
@donn are there open github issues for the "odd issues with TritonCTS"?  Are you going to make the default switch (I didn't make the meeting)?

donn:
@maliberty Not yet no. Most likely not- we're going to switch to simple CTS on the designs that have hold vios with TritonCTS.

PR #662, opened 2021-10-21
docker/cugr: add missing bzip2 dependency

oharboe:
Fixes build problem with docker.


donn:
Moot as of docker restructure

PR #663, opened 2021-10-21
cleanup: remove binaries from git repository

oharboe:


donn:
This will break the CI and make maintainers' jobs a lot harder. We don't rebuild each tool every time. Until we have an alternative method of caching the binaries somehow, the tarballs are here to stay.

donn:
Possible- the storage limits are a bit muddy. We're currently working with @hdl to have an all-docker images solution (they have funding from Google to store images in gcr.io), but that's gonna take a while to materialize.

oharboe:
> Possible- the storage limits are a bit muddy. We're currently working with @hdl to have an all-docker images solution (they have funding from Google to store images in gcr.io), but that's gonna take a while to materialize.

Yes. A docker only storage makes a lot of sense. Thanks for explaining your stop-gap solution, makes sense.

PR #686, opened 2021-11-01
Changes to device level LVS

d-m-bailey:
- Always include the `STD_CELL_LIBRARY` (previously defining `LVS_EXTRA_STD_CELL_LIBRARY` would cause `STD_CELL_LIBRARY` to be ignored)
- Added `LVS_EXTRA_SPICE` variable (functionally equivalent to `LVS_EXTRA_STD_CELL_LIBRARY` but makes more sense for analog or sram modules.)
- Flatten cells matching `*_example_*` during extract
- Flatten small cells during extract


donn:
@Manarabdelaty Mind taking a look at this? It's a bit out of my depth.

donn:
@d-m-bailey I took the liberty of updating the changes- can you take a look? If they match your intention, I'll merge.

d-m-bailey:
@donn Actually, this program needs to be modified a bit more with the most recent version of `lvs.tcl`, so I'd recommend holding off on the merge for a couple days. I'm working on a new PR and I'll explain the issues in more detail there.

Most importantly, the addition of `LVS_INSERT_POWER_PINS` processing breaks the `run_lvs_batch` procedure in `flow.tcl`.

Also only the flattening of cells with names containing `example` is only a partial solution. See issue #796

donn:
Gotcha. I'll wait.

PR #691, opened 2021-11-03
Update README.md

mattvenn:
Take @maliberty explanation in the skywater slack and add to the parameters on hold and setup slack.


donn:
Thanks, matt- in the future just add `[skip ci]` in the commit description of documentation-only changes so we can save some compute time. :)

donn (approved):

PR #700, opened 2021-11-10
Disable timing and routability driven modes when not wanted

antonblanchard:
In global placement, OpenLane assumes timing and routability driven modes
are disabled. This is not the case when calling the OpenROAD APIs
directly, so explicitly disable them.


donn (approved):
While this would indeed be made redundant by the other issue, I'm gonna go ahead and merge it for now anyway.

PR #702, opened 2021-11-11
OpenROAD Script Updates

### LINKED ISSUE #642: Enable the OR GUI
Please enable the OR GUI in your docker build of OR.  You will need to have Qt installed in your container.  See https://github.com/The-OpenROAD-Project/OpenROAD/blob/master/etc/DependencyInstaller.sh for the OR setup.

### LINKED ISSUE #697: rewrite or_replace.tcl to use global_placement command not low-level swig
### Description
In or_replace.tcl you should be using the global_placement command not mucking around with the swig layer.  The current code for PL_ROUTABILITY_DRIVEN is wrong and routability can't be turned off.  Please replace all this with a proper command invocation.

### Environment
latest

### Reproduction Material
any design

### Expected behavior
match OR

### Logs
None

donn:
* Rewrite or_replace.tcl

That script used a lot of internal swig commands with an unstable API/unpredictable behavior. Resolves #697.

* Expose `max_displacement` flag for detailed placement

This was added at some point as a replacement to the diamond search height thing and is necessary for macros.

cc @mguthaus 

* Add OpenROAD GUI/Tcl Readline

Resolves #642. Note that this addition will be intentionally undocumented and is intended more for people who absolutely want it.

MISC:

* CI fast fail off/always push image

The CI will report failures still, but will still publish an image and create a tag. Additionally, other tests won't be killed  on the first failure.

* Naming convention changes

  * `docker_build/` -> `docker`: Just is more consistent with literally every other codebase that uses Docker ever.

  * Removed `or_` prefix from OpenROAD scripts: redundant: They're all already under a folder called "openroad." The distinction no longer makes any sense as there are no separate scripts for the previously standalone utilities.


dependencies/centos-7/run_time.txt
```diff
@@ -16,8 +16,11 @@ perl-Switch
 python36u
 python36u-pip
 python36u-tkinter
+qt5-qtbase
+qt5-qtimageformats
```
maliberty, line 20: I think this needs to be in compile_time.txt too.

docker/openlane/Dockerfile
```diff
@@ -69,4 +69,6 @@ COPY ./.tclshrc /root
 
 WORKDIR $OPENLANE_ROOT
 
+RUN dbus-uuidgen --ensure
```
maliberty, line 72: Just curious what this is for?

scripts/openroad/replace.tcl
```diff
@@ -49,59 +49,41 @@ if { ! $free_insts_flag } {
...
 }
 
 if { $::env(PL_TIME_DRIVEN) } {
 	read_sdc $::env(CURRENT_SDC)
-	gpl::set_timing_driven_mode 1
 	read_verilog $::env(yosys_result_file_tag).v
+	lappend arg_list -timing_driven
 } else {
-	gpl::set_timing_driven_mode 0
+	lappend arg_list -disable_timing_driven
```
maliberty, line 71: This will generate a "-disable_timing_driven is deprecated." warning.  You can just drop it.

scripts/openroad/replace.tcl
```diff
@@ -49,59 +49,41 @@ if { ! $free_insts_flag } {
...
-	grt::add_layer_adjustment 3 $::env(GLB_RT_L3_ADJUSTMENT)
-	if { $::env(GLB_RT_MAXLAYER) > 3 } {
-	   grt::add_layer_adjustment 4 $::env(GLB_RT_L4_ADJUSTMENT)
-	    if { $::env(GLB_RT_MAXLAYER) > 4 } {
-	        grt::add_layer_adjustment 5 $::env(GLB_RT_L5_ADJUSTMENT)
-	        if { $::env(GLB_RT_MAXLAYER) > 5 } {
-	            grt::add_layer_adjustment 6 $::env(GLB_RT_L6_ADJUSTMENT)
-	        }
-	    }
-	}
```
maliberty, line 89: Where did these adjustsment go?

scripts/openroad/replace.tcl
```diff
@@ -49,59 +49,41 @@ if { ! $free_insts_flag } {
...
+	lappend arg_list -routability_max_inflation_iter 10
 } else {
-	gpl::set_routability_driven_mode 0
+	lappend arg_list -disable_routability_driven
 }
 
-if { !$::env(PL_SKIP_INITIAL_PLACEMENT) || $::env(PL_BASIC_PLACEMENT) } {
-    gpl::replace_initial_place_cmd
+if { $::env(PL_SKIP_INITIAL_PLACEMENT) } {
+	lappend arg_list -skip_initial_place
```
maliberty, line 83: Was it intentional to drop || $::env(PL_BASIC_PLACEMENT ?

dependencies/centos-7/run_time.txt
donn: Anything in run_time is also in compile_time.

docker/openlane/Dockerfile
donn: I think Qt needs it?

scripts/openroad/replace.tcl
donn: Yes, I noticed. I'm removing them.

scripts/openroad/replace.tcl
donn: Is there an equivalent set of commands for placement? Should I just keep the grt:: commands?

scripts/openroad/replace.tcl
donn: No, I'm restoring it.

scripts/openroad/replace.tcl
maliberty: set_global_routing_layer_adjustment

docker/openlane/Dockerfile
maliberty: Its not part of the OR build afaik.

donn: Yeah iunno, without this command it would tell me there's a missing uuid.

kareefardi (approved):

PR #706, opened 2021-11-12
fix to address issue with modified decap_12

jeffdi:
emergency fix to support mpw-3.  updates magic and open_pdks commit.


donn:
Recreated as #707.

PR #708, opened 2021-11-13
Update documentation to show RUN_SIMPLE_CTS defaults to off

antonblanchard:


donn (approved):
Oops. Thanks for catching that.

PR #712, opened 2021-11-14
Remove *_REPO and *_COMMIT defaults in Dockerfiles

antonblanchard:
It took me a while to realise we no longer specify commit IDs directly
in Dockerfiles, but now use dependencies/tool_metadata.yml. To make it
clearer, remove the defaults in the Dockerilfes.


antonblanchard:
The commit IDs are likely the most confusing, not sure if you want to keep the GIT URL defaults.

FYI: I only tested a rebuild of the openroad target with this patch.

donn (approved):

PR #713, opened 2021-11-14
Fix typo in run_resizer_design

antonblanchard:
When skipping design resizing, we incorrectly printed we were
skipping timing resizing.


donn (approved):
Thanks for catching that.

PR #716, opened 2021-11-16
Move fill insertion after detailed placement

antonblanchard:
I have a design that fails detailed placement just before global routing.
Looking closer, I'm not sure why we do fill insertion before the final
detailed placement step, because adding filler cells is just going to
make it impossible to fix any placement issues.

### DIFF
--- a/scripts/tcl_commands/routing.tcl
+++ b/scripts/tcl_commands/routing.tcl
@@ -383,12 +383,6 @@ proc run_routing {args} {
 		}
     }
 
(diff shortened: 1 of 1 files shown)


antonblanchard:
This fixes my issue, but could do with some broader testing.

donn:
This looks okay to me, @Manarabdelaty what do you think?

Manarabdelaty:
@donn Looks okay to me too

Manarabdelaty (approved):

PR #721, opened 2021-11-18
Non-fixed layer count for GLB RT adjustments

donn:
This change is to help people implement PDKs that may more have than six metal layers, which I'm going to go out on a limb and say, exist.

Remember kids, nothing is more permanent than a temporary solution.

* GLB_RT_LX_ADJUSTMENT variables consolidated into GLB_RT_LAYER_ADJUSTMENTS, a comma-delimited set of values

* GLB_RT_LX_ADJUSTMENT will continue to be accepted (and converted into GLB_RT_LAYER_ADJUSTMENTS) but display a deprecation warning, and will be removed in a later version of OpenLane.

* The GLB_RT_L1_ADJUSTMENT value from the PDK is now ignored.


kareefardi (approved):

maliberty:
Yes they exist :)  Nice change

PR #724, opened 2021-11-19
Fix sram install

Manarabdelaty:
- This should fix the SRAM installation for the open_pdks version we are using. 
- By default, the SRAM installation is disabled. To enable, `export INSTALL_SRAM=enabled` .

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -42,6 +42,12 @@ endif
 STD_CELL_LIBRARY ?= sky130_fd_sc_hd
 SPECIAL_VOLTAGE_LIBRARY ?= sky130_fd_sc_hvl
 IO_LIBRARY ?= sky130_fd_io
+INSTALL_SRAM ?= disabled
+
+OPEN_PDK_ARGS ?= ""
+ifeq ($(INSTALL_SRAM), enabled) 
+OPEN_PDK_ARGS ?= --enable-sram-sky130
+endif 
 
 ifeq ($(OPENLANE_IMAGE_NAME),)
 OPENLANE_TAG ?= $(shell python3 ./dependencies/get_tag.py)
@@ -135,7 +141,7 @@ build-pdk: $(PDK_ROOT)/open_pdks $(PDK_ROOT)/skywater-pdk
 		rm -rf $(PDK_ROOT)/sky130A) || \
 		true
 	$(ENV_COMMAND) sh -c " cd $(PDK_ROOT)/open_pdks && \
-		./configure --enable-sky130-pdk=$(PDK_ROOT)/skywater-pdk/libraries --enable-sram-sky130"
+		./configure --enable-sky130-pdk=$(PDK_ROOT)/skywater-pdk/libraries $(OPEN_PDK_ARGS)"
 	cd $(PDK_ROOT)/open_pdks/sky130 && \
 		$(MAKE) veryclean && \
 		$(MAKE) prerequisites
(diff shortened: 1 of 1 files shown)


donn (approved):

d-m-bailey:
Just a second @Manarabdelaty @donn . (adding @RTimothyEdwards )

Is there an anticipated change to `open_pdks`?

1. Here's the code from `open_pdks/scripts/configure`
```
    # Check whether --enable-sram-sky130 was given.
if test "${enable_sram_sky130+set}" = set; then :
  enableval=$enable_sram_sky130;
            if test "$enableval" == "yes" -o "$enableval" == "YES"; then
                { $as_echo "$as_me:${as_lineno-$LINENO}: Package 'sky130_sram_macros' will be installed automatically during make." >&5
$as_echo "$as_me: Package 'sky130_sram_macros' will be installed automatically during make." >&6;}
                export SKY130_SRAM_MACROS_PATH=../sources/sky130_sram_macros
            elif test "$enableval" == "no" -o "$enableval" == "NO"; then
                { $as_echo "$as_me:${as_lineno-$LINENO}: Disabling package 'sky130_sram_macros'" >&5
$as_echo "$as_me: Disabling package 'sky130_sram_macros'" >&6;}
                export SKY130_SRAM_MACROS_PATH=""
            else
                SKY130_SRAM_MACROS_PATH=$enableval
                { $as_echo "$as_me:${as_lineno-$LINENO}: Enabling package 'sky130_sram_macros' at $SKY130_SRAM_MACROS_PATH" >&5
$as_echo "$as_me: Enabling package 'sky130_sram_macros' at $SKY130_SRAM_MACROS_PATH" >&6;}
            fi
            SKY130_SRAM_MACROS_PATH=`realpath $SKY130_SRAM_MACROS_PATH`

else

            { $as_echo "$as_me:${as_lineno-$LINENO}: Package 'sky130_sram_macros' will not be installed." >&5
$as_echo "$as_me: Package 'sky130_sram_macros' will not be installed." >&6;}


fi
```
So `--enable_sram_sky130=yes` installs in the default directory `../sources/sky130_sram_macros`,
`--enable_sram_sky130=no` does not install sram macros,
and `--enable_sram_sky130=my_directory` installs in  `my_directory`.
So wouldn't `--enable_sram_sky130` attempt to install in the `open_pdks/scripts/..` directory? Is that what is wanted?
The previous version had `INSTALL_SRAM ?= disabled` which would try to install from `open_pdks/scripts/disabled/..` and silently fail.
Here's the log
```
Magic version 8.2 (or better) available on the system.
Install option: lef
Collecting files from /<dir>/openlane/pdks/open_pdks/scripts/disabled/../sky130_sram_macros/*/*.lef
```
Checking the directory gives
```
$ ls <dir>/openlane/pdks/open_pdks/scripts/disabled/../sky130_sram_macros
ls: cannot access '<dir>/openlane/pdks/open_pdks/scripts/disabled/../sky130_sram_macros': No such file or directory
```

2. Also, I don't think this works as you want. 
```
OPEN_PDK_ARGS ?= ""
ifeq ($(INSTALL_SRAM), enabled) 
OPEN_PDK_ARGS ?= --enable-sram-sky130
endif 
```
Won't this do nothing if `OPEN_PDK_ARGS` is already defined?
Maybe you intended `+=`
```
OPEN_PDK_ARGS ?= ""
ifeq ($(INSTALL_SRAM), enabled) 
OPEN_PDK_ARGS += --enable-sram-sky130
endif 
```
But then again, this will attempt to install the repo at `open_pdks/scripts/..`, which may not be where you want. This installation takes close to an hour or more IIRC. Are you sure you want it enabled by default?

So with the current changes, `export INSTALL_SRAM=my_directory` would no longer work, because it's not passed to the `configure` program. In order to get the previous functionality, the user would need to set the environment variable before the Makefile. `export OPEN_PDK_ARGS="--enable-sram-sky130=my_directory"` or for installation in the previous default directory `export OPEN_PDK_ARGS="--enable-sram-sky130=yes"`.


I'd recommend leaving the configure statements as they were, adding an explanation, and changing the default to `yes` (if you really want to) or `no` for the correct disabling.
```
# INSTALL_SRAM valid values: 
# yes - install in default location, 
# no - do not install
# <directory> - install from <directory>
INSTALL_SRAM ?= yes
```

What do you think?

RTimothyEdwards:
@d-m-bailey :  `--enable_sram_sky130=my_directory` means "get source files from my_directory", not "install into my_directory".  If you use "=yes", then it assumes that you do not have the SRAM library repository, and it clones it automatically from github.  If you already have cloned the repository and it exists somewhere in your filesystem space, then "=my_directory" tells open_pdks to use that location as the source and not to do its own clone of the repository.

Issue #727, opened 2021-11-23
Flow fails when using inverter example with sky130_fd_sc_hvl


hpretl:
### Description

Tried to run the stock `inverter` example, only changed the library in `config.tcl` to

`set ::env(STD_CELL_LIBRARY) "sky130_fd_sc_hvl"`

Run fails on step 2 with:

**Error: sta.tcl, 42 cannot read file  /usr/local/share/pdk/sky130A/libs.ref/sky130_fd_sc_hvl/lib/sky130_fd_sc_hvl__tt_025C_3v30.lib /usr/local/share/pdk/sky130A/libs.ref/sky130_fd_sc_hvl/lib/sky130_fd_sc_hvl__tt_025C_3v30_lv1v80.lib .**

Both files are available and accessible. Not sure if the trailing `.` causes the issue, but I also do not know where this is introduced.

### Environment
```
Python: v3.9.7
Kernel: Linux v5.13.0-20-generic
Distribution: ubuntu 21.10
Container Engine: docker v20.10.7
OpenLane Git Version: 2021.11.23_01.42.34
---
PDK Version Verification Status: OK
---
Git Log (Last 3 Commits)

commit e1883edfbcc8c7e96491158521522b7a66541745
Author: Manar <manarabdelatty@aucegypt.edu>
Date:   Mon Nov 22 07:59:18 2021 -0600

    Update resizer (#726)

commit 6245b0e675d0e6751bdfc5362c545efe29d1077c
Author: Manar <manarabdelatty@aucegypt.edu>
Date:   Fri Nov 19 13:38:50 2021 -0600

    Fix sram install (#724)
    
    Also disable by default

commit e0e6fbc1bd90d1c73aae19769d997da8c3c47512
Author: Manar <manarabdelatty@aucegypt.edu>
Date:   Fri Nov 19 12:30:08 2021 -0600

    Add option for enabling/disabling repair_tie_fanout (#722)
    
    * Add option for enabling/disabling repair_tie_fanout
    
    * add escape characters
    
    * Add option for enabling sta report power
```

### Reproduction Material

I am using the stock `inverter` example in the design folder.

### Expected behavior

Running without fail ?!

### Logs
```
```

d-m-bailey:
Possibly related to issue #683 which was supposed fixed by PR #714

PR #728, opened 2021-11-23
Proposed fix for #727.

hpretl:
Using the fix below the inverter example runs with the `sky130_fd_sc_hvl` lib.


donn (approved):
Looks decent, will merge when CI passes.

PR #730, opened 2021-11-23
Update openroad commit

Manarabdelaty:


donn:
https://github.com/The-OpenROAD-Project/OpenLane/pull/734 is a bit more thorough.

PR #731, opened 2021-11-23
assert existence of EXTRA_GDS_FILES

kareefardi:
Files that don't exist in the variable `EXTRA_GDS_FILES` are reported silently during generating the finishing gds. This commit quits the flow if a file in the variable doesn't exists before starting the flow itself.


donn (changes requested):
Throw an exception. Exit 1 breaks the flow without reporting anything.

kareefardi:
> Throw an exception. Exit 1 breaks the flow without reporting anything.

So `return -code error` ?

donn:
```tcl
		flow_fail
		return -code error
```

donn:
Where's the `flow_fail` mate

kareefardi:
> ```tcl
> 		flow_fail
> 		return -code error
> ```

Sorry. I didn't see this.

donn (approved):

PR #732, opened 2021-11-23
Fix typos in PL_RESIZER_REPAIR_TIE_FANOUT option

antonblanchard:


donn (approved):
Thank you for catching that...

PR #738, opened 2021-11-25
correctly set when INSTALL_SRAM=enabled OPEN_PDK_ARGS

kareefardi:
Properly append `--enable-sram-sky130` to `OPEN_PDK_ARGS`.


donn (approved):

d-m-bailey:
This is only a partial fix of the problems mentioned in https://github.com/The-OpenROAD-Project/OpenLane/pull/724#issuecomment-974502254

Issue #484, opened 2021-07-05
Provide detailed instructions for makefile to support docker connection to XServer


dralabeing:
Provide clear and detailed instructions for running GUIs within the docker- specifically for Xserver support. Merely, setting the DISPLAY variable is not sufficient.  The docker cmd needs a -e variable, for e.g 
docker run -v /tmp/.X11-unix:/tmp/.X11-unix -e DISPLAY=unix:1 -it --rm -v $(OPENLANE_DIR):/openLANE_flow [...]
This is necessary for remote machine operation and will save first time users a lot of time

A useful link: https://l10nn.medium.com/running-x11-applications-with-docker-75133178d090

leochand101:
Any updates on this ? I realized this issue in my first build as well, whats the intended use case ? I was trying to open the built gds on klayout to find the Xserver error. 
Should I build from source or is docker file image the typical use case in which case we do need to have X11 support.

dralabeing:
@donn Any updates on this?

donn:
This is not an intended use case for OpenLane. If people would like to view their designs in Klayout, the recommended approach is to install Klayout on their machine. We only use Klayout/Magic/etc in batch mode.

maliberty:
@donn why not just add a few lines of documentation and enable this for users?  In ORFS we have make targets to run klayout on the result.

donn:
Not opposed to it if someone can spare the manpower to do so, but given that Klayout has been a particular trouble spot, I... do not have the mental energy to add and test a new set of options. :)

maliberty:
@dralabeing do you want to make a PR for this?

vijayank88:
> Any updates on this ? I realized this issue in my first build as well, whats the intended use case ? I was trying to open the built gds on klayout to find the Xserver error.
> Should I build from source or is docker file image the typical use case in which case we do need to have X11 support.

Try this docker command while mount:
docker run -v /tmp/.X11-unix:/tmp/.X11-unix -e DISPLAY=unix:0.0 -it --rm -v $(OPENLANE_DIR):/openLANE_flow [...]

vijayank88:
> Any updates on this ? I realized this issue in my first build as well, whats the intended use case ? I was trying to open the built gds on klayout to find the Xserver error.
> Should I build from source or is docker file image the typical use case in which case we do need to have X11 support.

I tried in my machine below format working fine. Use below command while mount openlane:
[...]

PR #740, opened 2021-11-30
Allow openroad gui to run inside "make mount"

maliberty:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,10 +27,8 @@ DOCKER_OPTIONS += --memory=$(DOCKER_MEMORY)
 endif
 
 UNAME_S := $(shell uname -s)
-ifeq (1,$(DOCKER_DISPLAY))
 ifeq ($(UNAME_S),Linux)
-DOCKER_OPTIONS += -e DISPLAY=$(DISPLAY) -v /tmp/.X11-unix:/tmp/.X11-unix
-endif
+DOCKER_OPTIONS += -e DISPLAY=$(DISPLAY) -v /tmp/.X11-unix:/tmp/.X11-unix -v $(HOME)/.Xauthority:/.Xauthority --network host
 endif
 
 THREADS ?= 1
@@ -187,13 +185,9 @@ pull-openlane:
 mount:
 	cd $(OPENLANE_DIR) && \
 		docker run -it --rm \
-			--net=host \
 			-e PDK_ROOT=$(PDK_ROOT) \
 			-v $(OPENLANE_DIR):/openlane \
 			-v $(PDK_ROOT):$(PDK_ROOT) \
-			-e DISPLAY=$(DISPLAY) \
-			-v $(HOME)/.Xauthority:/.Xauthority \
-			-v /tmp/.X11-unix:/tmp/.X11-unix \
 			$(DOCKER_OPTIONS) $(OPENLANE_IMAGE_NAME)
 
 MISC_REGRESSION_ARGS=
--- a/Makefile
+++ b/Makefile
@@ -186,7 +186,15 @@ pull-openlane:
 .PHONY: mount
 mount:
 	cd $(OPENLANE_DIR) && \
-		docker run -it --rm -v $(OPENLANE_DIR):/openlane -v $(PDK_ROOT):$(PDK_ROOT) -e PDK_ROOT=$(PDK_ROOT) $(DOCKER_OPTIONS) $(OPENLANE_IMAGE_NAME)
+		docker run -it --rm \
+			--net=host \
+			-e PDK_ROOT=$(PDK_ROOT) \
+			-v $(OPENLANE_DIR):/openlane \
+			-v $(PDK_ROOT):$(PDK_ROOT) \
+			-e DISPLAY=$(DISPLAY) \
+			-v $(HOME)/.Xauthority:/.Xauthority \
+			-v /tmp/.X11-unix:/tmp/.X11-unix \
+			$(DOCKER_OPTIONS) $(OPENLANE_IMAGE_NAME)
 
 MISC_REGRESSION_ARGS=
 .PHONY: regression regression_test


maliberty:
This works for me and mostly addresses https://github.com/The-OpenROAD-Project/OpenLane/issues/484

maliberty:
@msaligane FYI

msaligane:
Yes, works for me too. Thanks @maliberty

donn (commented):
This is already supported by the current Makefile if you pass the environment variable DOCKER_DISPLAY=1.

mithro:
FYI - https://blog.yadutaf.fr/2017/09/10/running-a-graphical-app-in-a-docker-container-on-a-remote-server/

maliberty:
The DISPLAY_OPTION is missing the mapping of .Xauthority  so it doesn't work for me.  I don't see any reason for it to be an option (it's never harmful to have and non-obvious).  #484 was not updated with this info.  I don't see it in the docs.

I'll update this PR to handle the "ifeq ($(UNAME_S),Linux)" bit.

mithro:
See also references at https://hdl.github.io/containers/ug/index.html#_tools_with_gui and https://github.com/ghdl/docker/blob/master/USE_CASES.md#environments-with-guis from @umarcor.

donn:
> The DISPLAY_OPTION is missing the mapping of .Xauthority so it doesn't work for me. I don't see any reason for it to be an option (it's never harmful to have and non-obvious). #484 was not updated with this info. I don't see it in the docs.

As we've discussed in our private weekly sync, I had no intention to document this option because I don't want to go around supporting it. The .Xauthority thing is a perfect example: mine works fine without it, but yours won't. I also don't get why you specified `--network host`.

I'm not opposed to turning it on by default given your statement that it's never harmful, however.

donn (approved):

maliberty:
I expect the reason is that I ssh into another machine where I run the container and the DISPLAY is not local.

umarcor:
As explained in the references provided by @mithro above, I strongly recommend using [mviereck/x11docker](https://github.com/mviereck/x11docker) and [mviereck/runx](https://github.com/mviereck/runx) (on Windows) in order to run GUI applications in containers.

- In essence, x11docker is a MIT licensed bash script (a single file) that handles the issues you've been discussing about (DISPLAY, Xauthority, etc.).
- Moreover, x11docker supports xpra, xephyr, wayland, weston... See https://github.com/mviereck/x11docker/wiki/X-server-and-Wayland-Options.
- The author of x11docker is very interested in the security of using containers (rather poor by default), so x11docker provides multiple thoroughly engineered options to make (remote) usage of containers as secure as possible. 
- x11docker can be used as a CLI tool, providing a consistent interface regardless of the host OS or the specific X server. Alternatively, it has a GUI (https://github.com/mviereck/x11docker#gui-for-x11docker) which can be executed from a container.
- Using x11docker for building and/or running EDA tools with GUIs has been discussed and tested: mviereck/x11docker#201.

See https://joss.theoj.org/papers/10.21105/joss.01349 and https://github.com/mviereck/x11docker/wiki.

Disclaimer: I've been involved in testing x11docker on Fedora and on Windows, as I pushed for adding Windows support to x11docker in order to ditch my multi-platforms scripts/makefiles.

maliberty:
@umarcor It would be great if you can jump in with a PR implementing a better solution.  I just got tired of people saying it doesn't work and passing around workarounds on gitter/slack.

donn:
@umarcor @maliberty +1 - please do be advised though, if this changes our current container build system significantly or impedes headless operation in any form, it will not be merged.

The GUI support continues to be very much an at-your-own-risk-type deal.

umarcor:
@maliberty unfortunately I am not familiar enough with the use cases of the makefiles in this repo, and why/when do users need a GUI and when they do not. It seems that a single Makefile is used for multiple use cases, including running tests, unattended usage or interactive usage. Hence, I can only suggest how to proceed by looking at the current `mount` target. The following target should be equivalent:

```makefile
x11docker:
	cd $(OPENLANE_DIR) && \
		x11docker -i --user=0 $(X11DOCKER_OPTIONS) -- \
			-e PDK_ROOT=$(PDK_ROOT) \
			-v $(OPENLANE_DIR):/openlane \
			-v $(PDK_ROOT):$(PDK_ROOT) \
			$(DOCKER_OPTIONS) \
			-- $(OPENLANE_IMAGE_NAME) \
			bash
```

At the same time, remove from `DOCKER_OPTIONS` all the stuff related to DISPLAY, xauth, etc. (x11docker takes care of it).

X11DOCKER_OPTIONS can be used for providing custom options to x11docker. For instance, on MSYS2 I would use `X11DOCKER_OPTIONS=--runx --no-auth`, because I use VcxSrv as the X server.

@donn, the x11docker target should be completely unrelated to the build system, the execution of tests, or any other usage of the makefile. It is used for passing PDK_ROOT, OPENLANE_DIR, DOCKER_OPTIONS and OPENLANE_IMAGE_NAME to the container. Any other approach to pass those arguments would work. In fact, I believe that the x11docker target might be written in a different makefile, which includes the current one.

grant-h:
Note that on Ubuntu 20.04 `--security-opt apparmor:unconfined` is _also_ required to be added to `DOCKER_OPTIONS`. Otherwise you'll get issues like `audit: type=1107 audit(1680185873.107:2557): pid=1232 uid=103 auid=4294967295 ses=4294967295 subj=unconfined msg='apparmor="DENIED" operation="dbus_signal"` in `dmesg` when running KLayout which says:
```
OpenLane Container (4cd0986):/openlane$ klayout -e -nn $PDK_ROOT/sky130A/libs.tech/klayout/tech/sky130A.lyt \
>    -l $PDK_ROOT/sky130A/libs.tech/klayout/tech/sky130A.lyp \
>    ./designs/spm/runs/openlane_test/results/final/gds/spm.gds
process 13: The last reference on a connection was dropped without closing the connection. This is a bug in an application. See dbus_connection_unref() documentation for details.
Most likely, the application was supposed to call dbus_connection_close(), since this is a private connection.
  D-Bus not built with -rdynamic so unable to print a backtrace
```

PR #742, opened 2021-12-01
Report Generation Bugfixes

donn:
\+ Added `-override_env` option to flow.tcl to override certain environment variables on a per-run basis
\+ Added option to extract inverter cells into the metrics behind an env var (so as not to break compatibility)
\+ Klayout requirements added to runtime
~ Regression test outputs renamed, design report now in yaml
~ CI now runs on ubuntu-20.04 specifically, installs pyyaml
~ Fixed `make test`
~ Rewrote compare_regression_reports in click
~ Fixed compare_regression_design (typo made it not run, aliasing failures!!)
~ Changes to variable names: FINAL_TIMING_REPORT_TAG is now LAST_TIMING_REPORT_TAG, ran every STA
~ Fixed yosys gate type extraction
~ **Reverted OpenROAD back to the last known good version**

---
We'll have to investigate the OpenROAD issue that was aliased by my typo.


maliberty (approved):
@Manarabdelaty please look at this when you have a chance.  I'm just approving to enable @donn to move forward.

donn:
I'm going to wait for Manar to take a look until I merge this. But I do urge her to be quick about it. This is critical.

Manarabdelaty (approved):

PR #748, opened 2021-12-06
Update write_sdf

Manarabdelaty:
- Add the divider option and set it to `.` to match the gate level netlist
- Add `include_typ` option to enable using cvc for timing simulations


donn (approved):

donn:
@Manarabdelaty New OpenROAD has been merged, might wanna try again

PR #749, opened 2021-12-06
Add variable for enabling -post_cts_disable option

Manarabdelaty:
- Adding `-post_cts_disable` reduces the skew significantly for large designs (for example: mgmt_core)


donn (approved):
Yeah, looks fine. I recommend adding new variables instead of [info exists] in general as you did here, that's probably why the other PR failed CI.

PR #756, opened 2021-12-11
update OR version and update for API changes

maliberty:

### DIFF
--- a/dependencies/tool_metadata.yml
+++ b/dependencies/tool_metadata.yml
@@ -78,7 +78,7 @@
     make PREFIX=$PREFIX install
 - name: openroad_app
   repo: https://github.com/The-OpenROAD-Project/OpenROAD
-  commit: e9bd11509b1d907cbc3205e17f699ef05ae2181c
+  commit: 174a460d88e87f538ac36e26ef82d0bba0f4182c
   build: |
     mkdir -p ./build
     cd ./build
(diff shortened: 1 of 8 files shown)


maliberty:
@donn I can't tell exactly what is failing here - there is tons of output.  I don't know if these are real or not.  Any hints?

donn:
These are definitely real. What you should typically do is check a failing design's report at the end of each run:

IMG_0127

Sometimes, the failures are because the designs are failing when compared to an earlier benchmark, but most likely failures are just when OpenLane fails outright. In which case, you need to download the logs. You can't download them per-design because that's just how GitHub works, so what you can do is download all of the run_tarballs.zip file from the CI run:

IMG_0128

Inside the tarball for each run, there's a file called flow_summary.log. With the APU as an example, here's the error:

IMG_0126
 
In this scenario, it appears that there have been some regression in OpenROAD causing designs that were previously working to fail.

When updating OpenROAD, what I typically do is run `make fastest_test_set` locally first as a sanity check, then see which designs fail, then package tarballs and pass them on to you guys. I can do that tomorrow if you'd like.

mithro:
Donn, the documentation you just wrote sounds like a great starting point
for a "Debugging CI" page in the developer documentation or wiki!
(Hopefully something like development docs exists somewhere?)

I would just start with copying your comment verbatim and then spend **no**
more then 10 minutes cleaning it up.

**Only** put more effort into improving the page when you share a reference
to it.

donn:
Agreed, @mithro. I'll see what I can do there.

---

APU issue is https://github.com/The-OpenROAD-Project/OpenROAD/issues/1419, isolating the rest...

donn (approved):
Looks great!

PR #757, opened 2021-12-13
Redo Klayout

donn:
Klayout is no longer an "OpenLane tool" proper, now it's just a dependency.

This is owing to the fact that building it is a goddamn nightmare.


maliberty:
@mithro here is another vote for "building it is a goddamn nightmare".  ORFS adopted the same approach.

donn:
To be clear here, I'm referring to the build times, which are really something else. It's not sustainable for us to keep building them when they're pre-build on klayout.de.

Manarabdelaty (approved):

PR #764, opened 2021-12-16
Centralize Calling OpenROAD + Automatic Reproducibles

### LINKED ISSUE #736: Add mechanism to create OpenROAD reproducibles on failure automatically
### Prompt
I am physically aging trying to get the reproducible for https://github.com/The-OpenROAD-Project/OpenLane/issues/710.

### Proposal
`try_catch openroad` failures should be automatically detected, and if they happen, or_issue.py should be called to create a reproducible that can be uploaded immediately instead of having to rely on the user to package it or an OpenLane maintainer to do so then pass it on to OpenROAD.

donn:
All OpenROAD scripts are now called with `run_openroad_script`, a variant of `try_catch`, that upon detecting a failure will automatically invoke `./scripts/or_issue.py` and create a reproducible within the run directory.

Also tweaks to how yosys is invoked.

---

Resolves #736.

### DIFF
--- a/configuration/README.md
+++ b/configuration/README.md
@@ -20,8 +20,8 @@ These variables are optional that can be specified in the design configuration f
 
 | Variable      | Description                                                   |
 |---------------|---------------------------------------------------------------|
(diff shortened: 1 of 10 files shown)


maliberty:
Awesome!

kareefardi (approved):

PR #767, opened 2021-12-16
Move some layer variables to PDK, update io_place.py

### LINKED ISSUE #766: avoid li routing
### Description
OL should avoid routing on the li layer in detailed routing.   Add ""-bottom_routing_layer met1" to the droute.tcl invocation of detailed route (or whatever variable holds met1).

### Environment
Head of master

### Reproduction Material
Inspect the script

### Expected behavior
As above

donn:
This renames a number of layer configuration related variables, makes them use the layer name instead of an ordinal index, and moves them to the PDK.

By default, the values have changed slightly: the minimum routing layer is now `met1`. Users can still choose to route in the local interconnect manually. See RTimothyEdwards/open_pdks#200 for the exact values.

---

Resolves #766.


kareefardi:
@donn I would separate global and detailed routing minimum and maximum layers. From what I gathered, global router would "advise" the detailed router to avoid a minimum or maximum layer while the detailed router argument is more aggressive and would completely avoid the set layers and would fail if it couldn't route without that specified layer. It's just a suggestion tho I could be wrong.

donn:
I'm afraid I'm not following. What would be a situation where you'd globally want to route with some layers but not detailed route, or vice versa?

maliberty:
li routing is not strictly bad but we generally avoid it.

donn:
> li routing is not strictly bad but we generally avoid it.

Sounds like a candidate for disabled by default.

donn:
aight if y'all want this merged someone better approve it

kareefardi (approved):

PR #792, opened 2021-12-22
Jenkins: split build stage

vvbandeira:
[skip ci]

Signed-off-by: Vitor Bandeira <vitor.vbandeira@gmail.com>


maliberty (approved):
FYI - according to @donn this is likely to change again soon as they are moving images into dockerhub.

PR #812, opened 2021-12-28
Update run_designs.py Documentation

FaragElsayed2:
** -- designs option for run_designs.py doesn't valid anymore, you can also use threads option do multiple runs in a parallel.


donn (approved):

donn:
Thank you.

PR #813, opened 2021-12-28
Update README.md

FaragElsayed2:
** -- designs option for run_designs.py doesn't valid anymore, you can also use threads option do multiple runs in a parallel.


donn:
Manually merged into #812.

PR #814, opened 2021-12-28
Update OpenROAD SWIG to 4.0.1

donn:
+ Add llvm-devtoolset-7-clang to CentOS 7 dependencies, just in case (OpenROAD was being funny with GCC for a second there)
~ Updates OpenROAD SWIG to 4.0.1
~ Update scripts to support changes in newer versions of OpenROAD
~ Onus of building images moved to the OpenLane CI proper instead of the tool update flow, which now just basically edits one file

---
[ci ets]


donn:
Blocked on https://github.com/The-OpenROAD-Project/OpenROAD/issues/1480

kareefardi (approved):

PR #839, opened 2022-01-03
Improve Environment Survey

donn:
+ Issue survey will now tell users explicitly if their Python or Docker versions are out of date
+ Environment survey now runs on Python 3.3 or higher, so it can accurately report to users that their Python version is below 3.6
~ Installer moved to its own file
- Removed emoji from or_issue (was misguided, really)

---
This is mostly a change to external tooling that should have no impact on OpenLane, but out of an excess of caution, I'll let the CI run.

### DIFF
--- a/dependencies/env_info.py
+++ b/dependencies/env_info.py
@@ -19,7 +19,11 @@ import sys
 import json
 import platform
 import subprocess
-from typing import Tuple, Union, List, Optional
+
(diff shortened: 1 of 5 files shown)


donn:
Excess of caution was warranted :)

kareefardi (approved):

PR #840, opened 2022-01-03
Update deflef_utils.tcl

kareefardi:
non-empty body


donn (approved):
note to self: dont pair program with vim users

PR #846, opened 2022-01-04
Allow Including sky130A in the OpenLane Docker Image

donn:
+ Add a dependency system for tools, so some tools can be built first in local installs or have other repos and commits included in docker image builds
+ Add fully working `open_pdks` image
+ Add options to list containerized tools in `dependencies/tool.py`, with a `--no-pdks` flag to exclude open_pdks
+ Add libraries/sky130_fd_pr/latest to the full PDK build (was missing)
~ Isolated PDK building stuff into `./dependencies/pdk.mk`
~ Final merge no longer uses a tarball, just uses a good 'ol copy

---
Caveat is this will not be enabled by default or used by the CI for now. The reason is the PDK adds another 3 gigabytes to the final image and takes positively forever to build.

To use an internally built PDK, you need to export two variables:

```
export EXTERNAL_PDK_INSTALLATION=0
export NO_PDKS=0
```

Then you'd `make openlane` as normal.

To finally enable this, we'd need to rewrite make_timing to not take forever (probably in Rust/Swift/Go/whatever) and Open_PDKs would need to be updated so it gzips the liberty files. We'd also need to ensure that all tools can load .lib.gz files.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -58,8 +58,8 @@ SKYWATER_COMMIT ?= $(shell $(PYTHON_BIN) ./dependencies/tool.py sky130 -f commit
 OPEN_PDKS_COMMIT ?= $(shell $(PYTHON_BIN) ./dependencies/tool.py open_pdks -f commit)
 
 PDK_OPTS = 
-# ifeq ($(EXTERNAL_PDK_INSTALLATION), 1)
-ifneq ($(EXTERNAL_PDK_INSTALLATION), 1)
+EXTERNAL_PDK_INSTALLATION ?= 1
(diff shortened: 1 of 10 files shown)


donn:
fyi @olofk

olofk:
I haven't reviewed it in detail but it looks good as far as I can see. A comment on the liberty.py script is that I did some benchmarking yesterday and concluded that it's the `liberty_dict` function that takes most of the time.

Also, it is totally fine to run all these things without conda, which I think is just an insane extra layer to run a single python script

kareefardi (dismissed):

kareefardi (approved):

PR #854, opened 2022-01-09
Assorted Fixes To Error Reporting System

### LINKED ISSUE #850: Keep Track Of Warnings
Reported by @kareefardi 

### Prompt
Warnings are printed mid-flow, where they can easily be ignored.

### Proposal
Warnings should be also be stored in an individual log file and printed at the very end.

donn:
\+ Warnings now printed by flow_fail and at the end of noninteractive flow runs
\+ Created warnings.log, errors.log with all warnings and errors
\+ Get Tag now returns a dummy tag if you're on a branch that's not `main` or `master`
\~ Tarballing moved to run_tests.py
\~ Restored `-u` permissions to run_test.py.
\~ Version mismatches now cause the flow to fail, straight up
\~ flow_summary.log -> openlane.log
\~ Various warning messages updated
\~ Version checking moved to `prep`
\~ `return -code error` added to flow_fail, removed after all flow_fail invocations
\- Removed `--tar_list` from run_designs.py - do your own tarring

---
Resolves #850


kareefardi (commented):
Looks good. But we also need to capture warnings generated by the flow tools.

donn:
> Looks good. But we also need to capture warnings generated by the flow tools.

You can go ahead and do that if you'd like.

donn:
Now, what's interesting here is that the macro placement test runs absolutely fine locally but craps itself in CI. I wonder...

donn:
im losing my goddamn mind

donn:
So??? APPARENTLY??? IT WORKS FINE WITH ROOT PERMISSIONS??????

AND NOT THE UID/GID????

donn:
Why is this showing up NOW???

image

donn:
Ah, okay. This issue exists when the PDK and the run do not share the same permissions, which causes magic to not read some cells properly and then the DRC goes haywire. This wasn't an issue outside the CI because we ensure the permissions match, but it doesn't work when CI is root. Horrendous dockerism. However, on top of that, it's never been noticed because the Magic DRC check was one of those fabled `flow_fail`s that just forgot to `return -code error`.

What a mess.

maliberty (approved):

PR #863, opened 2022-01-13
Disabling hold timing violations fixing for inputs and outputs

donghl17:
We add an option `ECO_SKIP_PIN` to the ECO flow so that the ECO step is now able to skip hold timing violation fixing for IOs.

By setting `ECO_SKIP_PIN` to 1, the ECO flow will ignore hold time violations for IOs, and only fix the reg-to-reg hold violations.


donn:
When you get a PR merged, you do not continue working on a fork. You rebase off of master and do your changes there.

I've manually applied commit 5b80533 in a cleaner PR: #867.

PR #867, opened 2022-01-14
ECO Flow Fixes

### LINKED ISSUE #860: Disabling hold timing violations fixing for inputs and outputs
Reported by @shalan 

### Prompt
The ECO flow currently attempts to fix hold timing violations on inputs and outputs, which won't really work if there are timing constraints.

### Proposal
Show how to use the ECO flow to disable hold timing violation fixing for IOs. If the feature doesn't exist, it should be surfaced.

donn:
+ Add option to skip hold timing violation fixing for I/O pins
+ Properly expose ECO flow Tcl commands
~ Formatting changes

---
Resolves #860.


msaligane:
Thanks @donghl17 and @donn this LGTM! 

@donghl17 I believe you tested this on a few designs (AES, JPEG and Litex mgmt_core)?

donghl17:
@donn I just made an improvement of the eco flow. Could you please give me access to commit to your branch?

donn:
@donghl17 You've recieved access to the branch. Commit what you will. https://github.com/donn/openlane/invitations

donn:
Please be aware I'm not signing off on this unless the CI passes.

msaligane:
@donn I have reviewed with Haolin a few times and this runs through a few design. So I approve it.

donn:
@msaligane GitHub needs to know that too so I can merge :)

msaligane:
Ofcourse! Thanks @donn

donn:
Err, that is... I'm waiting on you. https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/approving-a-pull-request-with-required-reviews

donn:
Alright, well, if you don't care about this PR I see no reason why I should.

msaligane:
@donn I did not see your response...

msaligane (approved):

scripts/tcl_commands/routing.tcl
```diff
@@ -437,13 +437,13 @@ proc run_routing {args} {
 
     # if diode insertion does *not* happen as part of global routing, then
     # we can insert fill cells early on
-    if { ($::env(DIODE_INSERTION_STRATEGY) != 3) && ($::env(ECO_FINISH) ==1) } {
+    if { ($::env(DIODE_INSERTION_STRATEGY) != 3) && ($::env(ECO_ENABLE) ==0 | $::env(ECO_FINISH) ==1) } {
```
antonblanchard, line 455: Did you mean to use bitwise or here?

donghl17: Thanks. The issue has been fixed.

donghl17:
@donn Is there a way to check which part of the code fails the lint test? I just did lint test myself and found there is one error of "unable to import odb" and 40 conventions. I am wondering if the odb error occurs because I am not using the right pylint config, or I am supposed to solve all the conventions to pass the test.

donn:
We don't do pylint. You can see what linters we use here: https://github.com/The-OpenROAD-Project/OpenLane/runs/5222722298?check_suite_focus=true

donghl17 (dismissed):

scripts/gen_insert_buffer.py
```diff
@@ -81,7 +82,7 @@
             vio_name = drcSections[i].strip()
             report_end_str = re.search("min_report_end", vio_name)
             if report_end_str is not None:
-                print("LLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLL")
+                print("not complete report")
```
msaligane, line 85: @donghl17 Can you change this to something like: report is in incomplete

msaligane (changes requested):

scripts/gen_insert_buffer.py
antonblanchard: I noticed this message when testing. If we hit here, doesn't it mean we collected all the records and it is complete?

antonblanchard: Longer term, I also wonder if there is a better to parse format we can use, vs the somewhat brittle parsing of the text output.

msaligane: I think this is totally possible. The reports are generated directly out OpenROAD/OpenSTA, so we can have a command that does this without having to parse the reports (which are most likely subject to change). This should be up to the flow owners/maintainers to coordinate.

msaligane (approved):

PR #871, opened 2022-01-19
Global routing doesn't detect clock nets

antonblanchard:
I noticed that global routing reports 0 clock nets:

[INFO GRT-0019] Found 0 clock nets.

Moving the reading of the SDC file earlier fixes it:

[INFO GRT-0019] Found 980 clock nets.


donn (approved):

donn:
Thank you.

PR #872, opened 2022-01-19
Bring mergeLibs.pl from ORFS to OL to handle multiple Liberty files i…

maliberty:
…n yosys

Required to make asap7 work as it uses multiple .libs unlike skywater*

### DIFF
--- a/scripts/mergeLib.pl
+++ b/scripts/mergeLib.pl
@@ -1,0 +1,61 @@
+#!/usr/bin/env perl
+
+# This script is sourced from Brown (with slight modifications). It merges
+# several timing libraries into one.
+# ------------------------------------------------------------------------------
+
+use strict;
+use warnings;
+
+my $sclname = $ARGV[0];
+shift @ARGV;
+my $cnt = @ARGV;
+
+if($cnt>0){
+  process_header($ARGV[0]);
+  my $file;
+  foreach my $file (@ARGV) {
+      process_cells($file)
+  }
+  print "\n}\n";
+} else {
+  print "use: mergeLib.pl new_library_name lib1 lib2 lib3 ....";
+}
+
+
+sub process_header {
+  my $filename  = shift;
+  open(my $fh, '<', $filename) or die "Could not open file $filename $!";
+  while (<$fh>) {
+    if(/library\s*\(/) {
+      print "library ($sclname) {\n";
+      next;
+    }
+    last if(/^[\t\s]*cell\s*\(/);
+    print $_;
+  }
+  close($fh)
+}
+
+sub process_cells {
+  my $filename  = shift;
+
+  open(my $fh, '<', $filename) or die "Could not open file $filename $!";
+
+  my $flag = 0;
+  # cut the cells
+  while (<$fh>) {
+    #chomp $_;
+    if(/^[\t\s]*cell\s*\(/) {#&& $flag==0){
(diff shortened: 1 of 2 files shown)


arlpetergadfort:
Why not just use the yosys interface with abc to send in all the lib? It's supported by yosys so there is no need to merge lib files

donn (changes requested):
mergeLib.pl needs a copyright header

mithro:
Didn't we just get rid of all the Perl files?

donn:
Yeah, I'm not sure if asking for a python rewrite is a bridge too far though

maliberty:
> Why not just use the yosys interface with abc to send in all the lib? It's supported by yosys so there is no need to merge lib files

@arlpetergadfort  I'm glad to abandon this PR if there is a better solution.  What specifically are you suggesting?

maliberty:
> Yeah, I'm not sure if asking for a python rewrite is a bridge too far though

IMO, I don't think the language matters much - it is a simple script that wouldn't be any better in python.

donn:
Just for maintainability reasons. The less languages the better.

arlpetergadfort:
> > Why not just use the yosys interface with abc to send in all the lib? It's supported by yosys so there is no need to merge lib files
> 
> @arlpetergadfort I'm glad to abandon this PR if there is a better solution. What specifically are you suggesting?

@maliberty presumably this is to overcome the one liberty file at the abc calls below?
https://github.com/The-OpenROAD-Project/OpenLane/blob/8ad16ce75c753a8effaa92cd977d3a30c8cb4a40/scripts/yosys/synth.tcl#L293
https://github.com/The-OpenROAD-Project/OpenLane/blob/8ad16ce75c753a8effaa92cd977d3a30c8cb4a40/scripts/yosys/synth.tcl#L315

If you need to pass in additional liberty files just add `-liberty fileA -liberty fileB...` to the abc call to have abc read `fileA` and `fileB`. The interface used to no support this, but it's does now, so there is no need concatenate the liberty files.

maliberty:
@arlpetergadfort it is more indirect than that, it is the call from yosys to yosys-abc during:
12.1.1. Executing ABC.

So I have to convince yosys to pass the flag.

arlpetergadfort:
> @arlpetergadfort it is more indirect than that, it is the call from yosys to yosys-abc during: 12.1.1. Executing ABC.
> 
> So I have to convince yosys to pass the flag.

@maliberty The two lines I highlighted should have been the ones that call yosys-abc.  Maybe I'm missing something else, but the yosys / yosys-abc interface can pass multiple liberty files, so the merging step is should not be needed.

maliberty:
@arlpetergadfort dfflibmap appears to only take one .lib file as an argument to -liberty.

arlpetergadfort:
@maliberty I see, I don't have a good solution there, except to suggest it would be possible to separately specify the file for the DFFLIB (usually the same as the other libs, but here a separate file I assume).

maliberty:
@arlpetergadfort in asap7 there are multiple .lib.  I think all the flops are in one of them so we could have a separate variable for the ff lib.  I don't know if that is always true though.

mithro:
Probably worth seeing if you can get a change into Yosys to allow multiple options if abc supports it.

maliberty:
@mithro  abc doesn't support it in dfflibmap - I'm sure it could be fixed but I'm not going to spend time on it.

PR #873, opened 2022-01-19
Remove manual tech mapping of muxes and get abc to map them

antonblanchard:
The manual techmap of 2 and 4 input muxes is causing a few issues. Firstly
it hides the mux from abc, so it is unable to apply any fan out
modifications. Secondly it doesn't handle the inverting output 2 input mux.


donn (approved):
Looks reasonable.

Issue #869, opened 2022-01-17
Specifying FP_DEF_TEMPLATE crashes the flow.


armleo:
### Description
Specifying FP_DEF_TEMPLATE crashes the flow. Issue traced to: `apply_def_template.py`. The bug is that click converts arguments cases to lowercase, but the CLI def expects templateDEF and userDEF. Tried manually fixing this. But instead got another bug of the same nature in defutils. Didnt debug further, decided to report.

You might be interested, in this too. If I fix this bug and specify non existing path to DEF it crashes with FileNotFound but the log does not make sense. Maybe a sanity check in OpenLane's scripts can be added?
### Environment
```
Kernel: Linux v5.11.0-46-generic

Distribution: ubuntu 21.04

Python: v3.9.5 (OK)

Container Engine: docker v20.10.10 (OK)

OpenLane Git Version: 2022.01.13_01.51.43

pip:click: INSTALLED

pip:pyyaml: INSTALLED

pip:venv: INSTALLED

---

PDK Version Verification Status: OK

---

Git Log (Last 3 Commits)



c01e53a 2022-01-12T18:40:17-03:00 Fix for make pdk (#861) - Vitor Bandeira -  (HEAD -> master, tag: 2022.01.13_01.51.43, origin/master, origin/HEAD)

e33dd27 2022-01-12T20:23:47+02:00 [BOT] Update openroad_app (#818) - Openlane Bot -  ()

45cb357 2022-01-12T19:50:06+02:00 Global Routing Fixes & Enhancements (#856) - Donn -  ()
```

### Reproduction Material
Any design with specified FP_DEF_TEMPLATE crashes.
### Expected behavior
It does not crash.
### Logs
```
[INFO]: Changing layout from /openlane/designs/user_analog_project_wrapper/runs/RUN_2022.01.17_10.12.50/tmp/floorplan/3-initial_fp.def to /openlane/designs/user_analog_project_wrapper/runs/RUN_2022.01.17_10.12.50/tmp/floorplan/4-io.def

[INFO]: Applying DEF template...

[ERROR]: during executing: "openroad -python /openlane/scripts/apply_def_template.py --def-template /openlane/designs/user_analog_project_wrapper/macros/def/user_analog_project_wrapper.def /openlane/designs/user_analog_project_wrapper/runs/RUN_2022.01.17_10.12.50/tmp/floorplan/4-io.def"

[ERROR]: Exit code: 1

[ERROR]: Last 10 lines:

[...]

PR #880, opened 2022-01-22
Fixes apply_def_template.py crashing see #869

armleo:
Fixes apply_def_template.py crashing see #869.

Clicks passes arguments to CLI all lower case, however the CLI function expects clicks to be case insensitive. Crash report can be seen in the issue mentioned above.

Still DEF test does not pass, as the Donn's pull request fixing defutil.py call has not been merged yet.

### DIFF
--- a/scripts/apply_def_template.py
+++ b/scripts/apply_def_template.py
@@ -21,9 +21,11 @@ import argparse
 import subprocess
 
 @click.command()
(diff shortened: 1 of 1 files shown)


donn:
Add the defutil.py fix here because that PR is gonna be a while before it's merged.

armleo:
Looks like, even with your pull request's changes the PINS location is not properly copied. Should I still pull request the apply_def_tempalte with replace_pins without the path changed to odbpy?

Issue #881, opened 2022-01-22
[Idea, Q&A] Add issue regression tests


armleo:
### Prompt
Currently only perfomance benchmarks are done and no compatability/feature tests are done.

For example

1. if you try to run a design wtih FP_DEF_TEMPLATE specified, it crashes, see: https://github.com/The-OpenROAD-Project/OpenLane/issues/869.

2. There is also other issues caused by a lot of reworks in the python scripts like defutil.py.

3. There is cases where small die areas crash the flow (see: https://github.com/The-OpenROAD-Project/OpenLane/issues/879)

4. Hi-Z inouts are connected to tie-low cell and it's impossible to disconnect it.

5. Specifying USE ANALOG in LEF should force the router to ignore the pin, however the flow crashes stating that USE ANALOG is an invalid.

6. Placing an LEF cell which contains met3 causes the flow to place met3 blockages over the macro (I assume at least), which in turn causes some met3 pins to be inaccessible.

This is just a couple I noticed, probably there is more.
### Proposal
My proposal is to add a set of tests that cover issues that happened in the past. This will help to reduce the gap between 100% coverage of all options and current state of OpenLane.

I am open to discussion, and if OpenLane devs arent against it, I can even implement it with integration with current CI flow.

armleo:
For example: Specifying the def currently causes the flow to not crash (with applied relevant fixes to above mentioned issues), however the floorplan's pin and macro locations dont match the specified def. This was not the case with v0.23.

If I added my def_test as a part of bug regression testing, this would allow me to create a test script which compares the flows reports and defs/gds to what should have been in the output, but only checking for the relevant parts, like pin/macro locations.

Obviously this creates a little more work, but this will reduce the amount of issues at the end of the day and make OpenLane more production-ready.

donn:
100% agreed. If you can get something implemented, please, by all means.

armleo:
Let me know what you think about my idea of implementing this.

Modify flow.tcl to source checks/post_run_hooks.tcl which will do the checking. This will allow to add a just any forms of checks in the flow, without introducing too much new code.

Then for issues, we will have minimal designs which reproduce this issues and checks can be added to checks/post_run_hooks.tcl that make sure that issue that we had with the design is didn't comeback as result of code changes.


[...]

PR #882, opened 2022-01-24
Issue regression flow

armleo:
See: https://github.com/The-OpenROAD-Project/OpenLane/issues/881

This commit adds -run_hooks flag which allows to run python scripts after design was ran. Example design was added. All designs should pass after this pull request, except new def_test design should be added but not run. If you run this, it should not pass, as the def flow has issues, that should be soon resolved.


donn:
Looks okay- though it will fail Python linting. We require black formatting and flake8 to pass.

armleo:
Will update the formatting asap

donn:
Wait, why? If you don't want to resolve the conflicts just tell me and I'll do it

armleo:
@donn I created one branch with all changes and created a merged pull request. This is my first time contributing using Pull Requests, I didn't know how to do this properly.

donn:
Ah, that makes sense.

PR #886, opened 2022-01-26
More generic flow updates

maliberty:
+ Add DFF_LIB_SYNTH optional variable for dfflibmap
~ ioplacer now uses commands rather than internals
~ -no_lvs -no_drc -no_antennacheck turned into flags
~ Handle multiple liberty files and empty CELL_PAD_EXCLUDE


donn (approved):
LGTM, will merge when CI passes

PR #891, opened 2022-01-27
Issue regression flow and apply_def_template.py fix

armleo:
This commit adds -run_hooks flag which allows to run python scripts after design was ran. Example design was added. All designs should pass after this pull request, except new def_test design should be added but not run. If you run this, it should not pass, as the def flow has issues, that should be soon resolved. See #881

Fixes apply_def_template.py crashing see #869

### DIFF
--- a/designs/def_test/config.tcl
+++ b/designs/def_test/config.tcl
@@ -1,0 +1,32 @@
+# User config
+set ::env(DESIGN_NAME) def_test
+
+# Change if needed
+set ::env(VERILOG_FILES) [glob $::env(DESIGN_DIR)/src/*.v]
+
+# turn off clock
+set ::env(CLOCK_TREE_SYNTH) 0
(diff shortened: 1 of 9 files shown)


donn (approved):
Looks good, thank you so much for your contribution. Will merge when CI clears.

maliberty:
@donn this partially reverted my changes and should not have been approved.  See flow.tcl

armleo:
This is correct, will fix this as soon as I will get to this.

maliberty:
Thanks, I've fixed it with another PR

armleo:
I merged with wrong commit. This was my first time doing a PR, so I messed up.

maliberty:
> I merged with wrong commit. This was my first time doing a PR, so I messed up.

No worries it wasn't hard to fix

PR #895, opened 2022-01-29
Added missing list braces

Syndace:
Fun little bug that makes the code fail when MACRO_BLOCKAGES_LAYER contains an even number of elements LOL

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -16,7 +16,7 @@ jobs:
       - name: Check out Git repository
         uses: actions/checkout@v2
       - name: Install Linters
-        run: python3 -m pip install black flake8
+        run: python3 -m pip install black==21.12b0 flake8==4.0.1
       - name: Ensure Black Formatting
         run: black --check .
       - name: Lint with Flake8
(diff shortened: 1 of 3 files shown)


donn (approved):
Ah, well, our decision to use Tcl is paying off dividends I see. Thank you.

donn:
The lint issue is likely because your changes are based on an older version of OpenLane. I'll just ignore it for the merge.

donn (dismissed):

Syndace:
The changes are based on the most recent commit of the master branch, so an older version can't be the cause for linting issues. I assume one of the Python linters that your pipeline uses has been updated in the last two days and complains about something that was okay in the previous version, since you are not pinning the versions of the linters as far as I can see.

donn:
wow good start to the week

donn (dismissed):
Thanks for noticing the linter thing. I'll update black sometime a little less weekend.

donn (approved):

Syndace:
(may I suggest doing `python3 -m pip install -r requirements_dev.txt` instead, single source of thruth and stuff, sorry if I'm annoying :D)

donn:
@Syndace I did have a requirements_lint.txt at some point but it got lost in merge hell at some point. I'll bring it back when I update black.

PR #896, opened 2022-01-30
added pdn and tap halo distinction

marwaneltoukhy:
Both tap halo and pdn halo were mapped to the same global variable, which did not make sense.
Updated the relevant documentation.


donn:
Ensure that you invoke `handle_deprecated_config`.

donn (approved):

Issue #892, opened 2022-01-27
DEF pin locations are not applied properly


armleo:
### Description
DEF pins locations are not used FP_DEF_TEMPLATE. Also the location of the component is not used from DEF. This might cause confusion, as FP_DEF_TEMPLATE's names assume that components locations are also used, but there is no proper documentation regarding the priority over manual macro placement.

Keep in mind #891 DOES NOT fix this. It just resurfaces it. If you use my branch you will see the issue if you run `./flow.tcl -design def_test -run_hooks`. It will run the def_test and the generated locations of the pins will not match the DEF.
### Environment
```
Kernel: Linux v5.11.0-49-generic

Distribution: ubuntu 21.04

Python: v3.9.5 (OK)

Container Engine: docker v20.10.10 (OK)

OpenLane Git Version: issue_regression_flow-dev

pip:click: INSTALLED

pip:pyyaml: INSTALLED

pip:venv: INSTALLED

---

PDK Version Verification Status: OK

---

Git Log (Last 3 Commits)



11bb94c 2022-01-28T01:00:25+04:00 Fixed flake8 complaining about tech variable in def_test/hooks/post_run.py - armleo -  (HEAD -> issue_regression_flow, origin/master, origin/HEAD)

dc5a5c2 2022-01-28T00:57:23+04:00 Fixed formatting - armleo -  ()

76273b6 2022-01-28T00:50:21+04:00 Fixed formatting - armleo -  ()
```

### Reproduction Material
See pull request above.
### Expected behavior
DEF's pin locations should be used properly. In v0.23 this was done properly def_util.py merge broke this. Or it might have happened in the past but resurfaced right now.
### Logs
Not relevant

armleo:
I am trying to fix defutil.py

Does anybody know how to define new Pins using opendb? Or maybe any documentation regarding the python API of OpenDB?

armleo:
Figured out, the integration tests show how its done, and the API is the same in python.

Currently trying to copy all the pins from one def to another, while filtering the ones that are in the first def.

armleo:
Currently with the Pull request accepted this has been partially resolved. However multiple pins are still not handled properly. Please don't close this, so I can fix this.

donn:
I have no intention to, don't worry.

armleo:
I am no longer working on the fix for this as I have no time.

donn:
That's totally fine! Thank you for telling me and not leaving us hanging.

[...]

PR #902, opened 2022-02-03
Add Issue-Specific CI

armleo:
Partial solution for #892 . The DEF pin locations should now be applied.

 Current implementation does not properly handle multipins with the same name, like in caravel user analog project (which is my current target, to get working). Currently working on implementing this, to solve def_test issue.

@donn I don't understand on how to include the def_test to regression designs. Any suggestions?


run_designs.py
```diff
@@ -32,7 +32,7 @@
 @click.option("-c", "--config_tag", default="config", help="Configuration file")
 @click.option("-r", "--regression", default=None, help="Regression file")
 @click.option("-t", "--tag", default="regression", help="Tag for the log file")
-@click.option("-j", "--threads", help="Number of designs in parallel")
+@click.option("-j", "--threads", default=4, help="Number of designs in parallel")
```
donn, line 35: Why?

scripts/apply_def_template.py
```diff
@@ -55,21 +56,30 @@ def remove_power_pins(DEF):
...
-            "/dev/null",
-            userDEF,
-            templateDEF,
-        ],
-        stderr=subprocess.PIPE,
-    )
+    shutil.copy(userDEF, f"{userDEF}.user.template.tmp")
+
+    try:
+        print(
```
donn, line 62: Why not just `subprocess.check_call` if you want to spit out the output anyway?

Additionally, defutil.py is not supposed to generate much in terms of output…

scripts/defutil.py
```diff
@@ -26,12 +30,24 @@ def cli():
...
         self.sites = self.lef.getSites()
+
         self.df = odb.read_def(self.db, def_in)
         self.block = self.db.getChip().getBlock()
         self.rows = self.block.getRows()
         self.dbunits = self.block.getDefUnits()
+        self.nets = self.block.getNets()
+
+
+# OdbReader("designs/def_test/runs/RUN_2022.01.30_12.32.26/tmp/merged.lef", "designs/def_test/runs/RUN_2022.01.30_12.32.26/results/final/def/def_test.def")
```
donn, line 44: No commented out giant test code blocks please.

scripts/defutil.py
```diff
@@ -152,18 +168,181 @@ def merge_pins(output, input_lef, def_one, def_two):
         f.write(merge_item_section("PINS", def_one_str, def_two_str))
 
 
-@click.command("replace_pins")
+# openroad -python scripts/defutil.py replace_pins -o output.def -l designs/def_test/runs/RUN_2022.01.30_12.32.26/tmp/merged.lef designs/def_test/runs/RUN_2022.01.30_12.32.26/tmp/floorplan/4-io.def designs/def_test/def_test.def
+
+
+@click.command(
```
donn, line 174: `click` idiomatically uses a docstring to print help for things.

scripts/tcl_commands/floorplan.tcl
```diff
@@ -254,6 +254,7 @@ proc apply_def_template {args} {
 	if { [info exists ::env(FP_DEF_TEMPLATE)] } {
 		puts_info "Applying DEF template..."
 		try_catch $::env(OPENROAD_BIN) -python $::env(SCRIPTS_DIR)/apply_def_template.py\
+			-l $::env(MERGED_LEF) \
```
donn, line 257: Don't mix long and short flags in the same invocation.

donn (changes requested):
Looks okay, just have a couple concerns:

donn:
> @donn I don't understand on how to include the def_test to regression designs. Any suggestions?

You *can* add it to `.github/test_sets/fastest_test_set`, but it's going to compare benchmarks. Benchmarks that won't exist.

I think we need a separate CI step for things that aren't just running an entire design.

armleo:
Will do, both suggestions and the CI.

run_designs.py
armleo: If -j is not passed, then run designs errors out cryptic message stating that empty string can't be converted to integer in line 118. Maybe should just set it to one.

scripts/apply_def_template.py
armleo: Will replace with check_call.

I think we should output the log, as this implementation, while solving def pin locations issue, will crash when strange input is supplied (like broken def). The log will help us debug this kind of issues. But I think, I will add an option to write it to some logfile instead of stdout.

scripts/defutil.py
armleo: The first line for ODB Reader should be removed.

Second line with openroad, I think I will leave, as it suggests a way to run this script properly, without launching the full flow.

scripts/tcl_commands/floorplan.tcl
armleo: Will do

scripts/defutil.py
armleo: Will fix.

run_designs.py
donn: That's more reasonable, yeah.

scripts/apply_def_template.py
donn: Makes sense.

scripts/defutil.py
donn: Add it to the docstring. Not in a comment.

armleo:
Will work on this soon. Also will fix this issue: #912 . If I forget, please remind me.

donn:
Reminder.

armleo:
Please don't merge this yet.

I am trying to figure out a way how to: Properly both log a warning in the log file AND put a warning in the run's stdout. Whatever I do, the stdout does not appear at the output of the run.

I will also add the issue regression flow CI scripts + will add the infrastructure to do error testing. Like to test if missaligned pin will put an error I want it to put.

armleo:
Currently:

Gives warning for missaligned pins, because of #912 .
Fixed #892 as the apply def template has been rewritten
Adds a issue_892_def_test test case
Adds issue_912_* test case. Which should (as expected) fail with warning.

Right now: Working on run_issue_regressions.py, which will run the issue_912_ test design and then run issue_912_ issue_regression_checks (not hooks). This will allow to test for the log to contain the warning and also test for return code to be not zero ( == not successful), as the error has been raised.

armleo:
Currently, I am stuck, as I don't know how to configure GitHub CI. This is ready to be pulled, I will add the GitHub CI next Sunday.

armleo:
Resolves following issues:
#912 warning is working
#892 def pin locations are not being applied
#881 Issue regression flow

Once CI finishes successfully, this can be merged

armleo:
Okay. I fixed the CI issue. This pull request is ready and can be merged. Let me know if there is things you want to be changed.

donn (approved):
This is really, really good work. Thank you.

PR #904, opened 2022-02-04
More ASAP7 Tweaks

donn:
+ Added GCD Design to test ASAP7
+ Added RUN_LVS variable to control running the LVS step
~ Allow running OpenLane without git repo cloned (sort of)
~ Updated Klayout Scrot, DRC to complain about a missing mr deck if Klayout is set as the primary signoff tool
~ Various Tcl Formatting

Co-authored-by: Matt Liberty <mliberty@eng.ucsd.edu>

---

`PDK=asap7 STD_CELL_LIBRARY=asap7sc7p5t ./flow.tcl -design gcd -no_drc`


maliberty (approved):
It looks fine to me.  I wonder if we can arrange to not need -no_drc when there is only klayout and no drc setup for it (which is all PDKs in ORFS).  I'll approve in case you want to tackle that separately.

donn:
That's a separate discussion yeah. Is there no Klayout DRC deck…?

maliberty:
> That's a separate discussion yeah. Is there no Klayout DRC deck…?

No there isn't

donn:
s-so how do you like. do drc

maliberty:
commercial drc only

PR #916, opened 2022-02-13
CVC Script Fixes

donn:
~ awk scripts isolated into their own files- why are we using awk, again?
~ various bugfixes/corner cases in run_lef_cvc

---

(cc @d-m-bailey sanity check would be appreciated)


d-m-bailey (commented):
Looks good.

You could simplify line 72 in `cvc.tcl`
`$::env(SCRIPTS_DIR)/cvc/$::env(PDK)` -> `$::env(CVC_SCRIPTS_DIR)`

marwaneltoukhy (approved):

kareefardi (approved):

PR #917, opened 2022-02-13
Update Magic, Open_PDKS

### LINKED ISSUE #898: Update netgen and magic
There have been some big improvements to netgen performance over the last month, and it would be good to get them into openlane. Magic has had a number of fixes too, including for an LVS issue I'm hitting quite a lot on openlane.

donn:
magic -> 7d601628e4e05fd17fcb80c3552dacb64e9f6e7b
open_pdks -> 7519dfb04400f224f140749cda44ee7de6f5e095

---

Resolves #898.

### DIFF
--- a/dependencies/tool_metadata.yml
+++ b/dependencies/tool_metadata.yml
@@ -35,7 +35,7 @@
   in_install: false
 - name: magic
   repo: https://github.com/rtimothyedwards/magic
-  commit: 6ecd077ab4e8a7b6ccd8922954cd56862a8d4c05
+  commit: 7d601628e4e05fd17fcb80c3552dacb64e9f6e7b
   build: |
     ./configure --prefix=$PREFIX $MAGIC_CONFIG_OPTS
     make clean
@@ -79,7 +79,7 @@
 - name: klayout
   repo: https://github.com/KLayout/klayout
   commit: 428d0fe8c941faece4eceebc54170cc04d916c03
-  build: ''
+  build: ""
   in_install: false
 - name: openroad_app
   repo: https://github.com/The-OpenROAD-Project/OpenROAD
@@ -106,4 +106,3 @@
   build: ""
   in_install: false
   in_container: false
-
--- a/dependencies/tool_metadata.yml
+++ b/dependencies/tool_metadata.yml
@@ -35,7 +35,7 @@
   in_install: false
 - name: magic
   repo: https://github.com/rtimothyedwards/magic
-  commit: 0bb6ac1fa98b5371c73156b6e876925397fb7cbc
+  commit: 6ecd077ab4e8a7b6ccd8922954cd56862a8d4c05
   build: |
     ./configure --prefix=$PREFIX $MAGIC_CONFIG_OPTS
     make clean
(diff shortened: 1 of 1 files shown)


proppy:
@donn do you have more details about the LVS missmatch?

I'm getting something like this on the `inverter` example: https://gist.github.com/proppy/2727a9de4c190d37f720b347f99df848

donn:
That is precisely what I'm getting.

antonblanchard:
Interesting, I've been using GDS based LVS, so haven't seen this. For some reason fill and decap cells are not getting collapsed by netgen. Might be a magic extraction issue:

```XFILLER_23_1661 vssd1 FILLER_23_1661/VNB vccd1 vccd1 sky130_fd_sc_hd__decap_12
XFILLER_27_417 vssd1 FILLER_27_417/VNB vccd1 vccd1 sky130_fd_sc_hd__decap_12
XFILLER_67_597 vssd1 FILLER_67_597/VNB vccd1 vccd1 sky130_fd_sc_hd__decap_12
XFILLER_82_545 vssd1 FILLER_82_545/VNB vccd1 vccd1 sky130_fd_sc_hd__decap_12
```

It looks like the bulk terminal isn't getting hooked up.

antonblanchard:
Looking at one decap cell in the magic extract:

```
use sky130_fd_sc_hd__decap_6 FILLER_211_1785 1 0 165324 0 -1 117504
merge "FILLER_211_1791/VPB" "FILLER_211_1785/VPWR"
merge "FILLER_211_1785/VPWR" "FILLER_211_1785/VPB"
merge "FILLER_211_1785/VPB" "FILLER_210_1789/VPWR"
merge "FILLER_211_1791/VGND" "FILLER_211_1785/VGND"
merge "FILLER_211_1785/VGND" "FILLER_210_1789/VGND"
```

I don't see any connection for the VNB pin

donn:
@RTimothyEdwards Thoughts?

RTimothyEdwards:
@donn :  Yes, I asked you for layout and you said you were busy but would get the layout to me. . .

antonblanchard:
Hi @RTimothyEdwards here's what I was testing with (user_proj_wrapper out of caravel): 
[bad-extract.tar.gz](https://github.com/The-OpenROAD-Project/OpenLane/files/8075201/bad-extract.tar.gz)

antonblanchard:
I put both GDS and DEF in the tarball

RTimothyEdwards:
I extracted your design and got a correct netlist.

antonblanchard:
I just updated my version of magic and extraction does look correct. I have no idea why the installed version was failing (a build from a week ago) and I haven't successfully found a failing version after building a number of commits around that time.

antonblanchard:
@RTimothyEdwards I think I understand what is going on:

[bad-extract-2.tar.gz](https://github.com/The-OpenROAD-Project/OpenLane/files/8084721/bad-extract-2.tar.gz)

Good spice output:
```
export MAGTYPE=mag
magic -noconsole -dnull -rcfile $PDK_ROOT/sky130A/libs.tech/magic/sky130A.magicrc spice.tcl
```

Bad spice output:
```
export MAGTYPE=maglef
magic -noconsole -dnull -rcfile $PDK_ROOT/sky130A/libs.tech/magic/sky130A.magicrc spice.tcl
```

With the good output looking like:
```
Xfill_0_58 VGND VPWR VGND VPWR sky130_fd_sc_hd__decap_12
Xfill_0_36 VGND VPWR VGND VPWR sky130_fd_sc_hd__decap_12
Xfill_0_25 VGND VPWR VGND VPWR sky130_fd_sc_hd__decap_12
```

and the bad:
```
Xfill_0_58 VGND fill_0_58/VNB VPWR VPWR sky130_fd_sc_hd__decap_12
Xfill_0_36 VGND fill_0_36/VNB VPWR VPWR sky130_fd_sc_hd__decap_12
Xfill_0_25 VGND fill_0_25/VNB VPWR VPWR sky130_fd_sc_hd__decap_12
```

antonblanchard:
The comment in Openlane says:
```
        # the following MAGTYPE has to be maglef for the purpose of LVS
        # otherwise underlying device circuits would be considered
        set ::env(MAGTYPE) maglef
```

antonblanchard:
That may be true, but it shows that maglef based extraction is broken. When I pulled my test cases out, I wasn't setting the variable, and getting mag based extraction.

RTimothyEdwards:
@antonblanchard :  I found that the extraction handles substrate differently for abstract views.  Basically it ignores the substrate for abstract views, which is based on an understanding of abstract views that they don't define any well or substrate layers.  That's not true for LEF 5.7 and up, and is not true of the sky130 standard cells.  So I have just pushed a change to magic that comments out the conditional that checks for whether a cell is a LEF-based abstract view and ignores the substrate.  I checked against some layouts from qflow using OSU035 (which are LEF views that don't have wells) and it gave the correct extraction, so I think this is properly corrected now.

antonblanchard:
Thanks @RTimothyEdwards, I've tested this and it fixes the issues I was seeing.

donn:
Thank you @RTimothyEdwards, @antonblanchard and @proppy for your assistance with this one. It's been weird.

kareefardi (approved):

PR #919, opened 2022-02-13
Make Klayout an OpenLane Tool Proper, Update to Klayout 0.27.7

donn:
+ Klayout is now a proper OpenLane tool

---

This PR gets rid of the last x86-64 exclusive package in OpenLane.

### DIFF
--- a/docker/klayout/build_klayout.sh
+++ b/docker/klayout/build_klayout.sh
@@ -1,0 +1,47 @@
+#!/bin/bash
+set -e
+set -x
+
+bininstdir="./tmp/bin" 
+builddir="./tmp/build" 
+
+# destination folders
+bindir="/build/bin"
+libdir="/build/lib/klayout"
+
+distpackdir="/build/lib/python3/dist-packages"
+pylibdir="$distpackdir/klayout"
+
+# clean bin directory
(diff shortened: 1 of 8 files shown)


kareefardi (approved):

docker/klayout/build_klayout.sh
```diff
@@ -0,0 +1,52 @@
...
+pylibdir="$distpackdir/klayout"
+
+# clean bin directory
+rm -rf $bininstdir
+
+sed -i 's/-O2 -g/-O2/' /usr/lib64/qt5/mkspecs/linux-g++/qmake.conf
+
+# do the actual build
+./build.sh\
+    -qt5 \
```
proppy, line 22: curious if openlane actually depends on the QT bits as part of the flow? (looks like klayout has some support to be built without it https://github.com/KLayout/klayout/blob/0132ad08f7623e9fe39d614dae7134814a404d33/build.sh#L472 but we'd have to investigate how "functional" it is without it)

donn: Some people choose to use the GUI by exposing their X11 to the Docker container.

Plus there's the screenshot feature that's admittedly disabled by default but does require a GUI

PR #924, opened 2022-02-15
Remove catch around PDN environment variables

antonblanchard:
It should be safe to remove these since they have been in the PDK
for quite a while.


antonblanchard:
I noticed this while looking for something else... I presume it's safe but could be missing the reason why it was added.

donn (approved):

PR #926, opened 2022-02-16
Add mount of ~/.openroad so the user GUI settings are available insid…

maliberty:
…e docker


donn:
Does this work fine if you don't have an ~/.openroad file?

maliberty:
I does 'work' but it creates a .openroad directory so I just made it conditional.

donn (approved):

PR #944, opened 2022-02-21
Use most recent STA file in ECO flow

antonblanchard:
glob returns an unsorted list of filenames, so create a loop that grabs
the mtime of all matching files and finds the newest one.


msaligane:
@antonblanchard  Yes this fixes the issue! Thanks!
We tried this PR on multiple designs and it works fine.

PR #946, opened 2022-02-21
Always call set_propagated_clock after CTS

antonblanchard:
Whenever we read an SDC file after CTS, be sure clock latencies are
taken into account by running set_propagated_clock [all_clocks]


donn (approved):
makes sense

PR #962, opened 2022-02-26
Apply fan out and max transition parameters to ORFS synth

antonblanchard:


antonblanchard:
FYI The abc buffering defaults are for a fan out of 10 and a slew of 100 picoseconds.

maliberty (approved):
This change seems fine but I do separately question

if {![info exist ::env(SYNTH_MAX_TRAN)]} {
	set ::env(SYNTH_MAX_TRAN) [expr {0.1*$clock_period}]

I suspect abc will use the liberty based value and there is no obvious relationship between clock period and max slew.  I would make the default to be unspecified instead.

antonblanchard:
> This change seems fine but I do separately question
> 
> if {![info exist ::env(SYNTH_MAX_TRAN)]} { set ::env(SYNTH_MAX_TRAN) [expr {0.1*$clock_period}]
> 
> I suspect abc will use the liberty based value and there is no obvious relationship between clock period and max slew. I would make the default to be unspecified instead.

ABC is doing this:

```
    pPars->Slew          = pAbc->pLibScl ? Abc_SclComputeAverageSlew((SC_Lib *)pAbc->pLibScl) : 100;
```

After loading the sky130 liberty file, it's reporting a target slew of 53 picoseconds:

```
	-S <num> : target slew in pisoseconds [default = 53]
```

I'm not entirely sure what a reasonable value is here. sky130 seems to have a max slew in the range of 1.5ns, so abc driving towards 53ps feels too low to me.

maliberty:
Its not clear to me how the "target slew" is related to the max slew.  The abc docs are wanting here...

antonblanchard:
I just did some testing, and on a few simple designs abc doesn't seem very sensitive to any value of -S (even extreme ones). I'll remove that part of the above patch for now.

maliberty (approved):
It's still used in abc_fine_tune but maybe it doesn't matter.

PR #965, opened 2022-02-27
Fix some STA clock issues

antonblanchard:


antonblanchard:
@maliberty @donn before this patch we run final STA with ideal clocks, unless the SDC file has `set_propagated_clock` in it. That seems bad since it could hide hold violations in a design.

donn (approved):

donn:
@shalan fyi

maliberty:
@antonblanchard I find this a bit confusing.  You seem to want to make propagated clocks on by default (using unset_propagated_clock in many places) but then you don't make it the default and require a -post_cts flag.  It seems like you would want a -pre_cts flag.  This change doesn't seem to cover eco.

antonblanchard:
Makes sense, I'll flip the polarity and use `-pre_cts`.

One thing that confused me is there are two completely different ways to call the same STA code:

- Standalone:
```
run_sta -post_cts -log $::env(routing_logs)/parasitics_sta.log
```
- Inline:
```
set ::env(RUN_STANDALONE) 0
source $::env(SCRIPTS_DIR)/openroad/sta.tcl
```

The standalone version is only used in two places. Synthesis, which makes sense, since we are in yosys and need to call out to openroad. It's also called for our final STA, and I'm not entirely sure why. It does allow the STA output to be placed into its own log file, but is that enough of a reason?

Calling it directly means the caller is responsible for correctly setting up the SDC parameters, and we wouldn't have this issue.

maliberty:
idk, @donn any comments?

donn:
Well, sometimes it's a step and sometimes (most of the time) it's run as part of another step. The final sta is treated as a standalone step independent of routing iirc.

antonblanchard:
I created PR #969 to flip the polarity

PR #969, opened 2022-02-28
Reverse polarity of run_sta -post_cts to run_sta -pre_cts

antonblanchard:


maliberty (commented):
Seems ok to me but I'll let @donn approve

donn (approved):

PR #972, opened 2022-03-01
ECO flow inserts buffers backwards

antonblanchard:
We currently connect the output of the FF to the output of the delay buffer.


antonblanchard:
@donghl17 @msaligane This could do with a close review. Does it look right to you?

msaligane:
@antonblanchard Good catch! We can see that as well in our netlist...

msaligane (approved):

donghl17:
@antonblanchard Thanks for fixing it! The delay is added to the circuit anyway so it's hard to notice. I did check this thing before actually, but something messed up when we tweak the code later.

PR #978, opened 2022-03-03
Multi-Process Corner RCX

### LINKED ISSUE #1020: Save SPEF file for all corners
I'm doing STA of a design with macros, so need the SPEF files for all corners. We could either save all of them by default, or else add an option.

donn:
+ Added optional multiple **process corner** technology lef support (min/max/nom), in addition to the already existing multiple **timing corner** support (ss/tt/ff). This will also be our nomenclature going fwd.
+ Parasitics extraction now run at all the process corners, and multi-corner analysis is run at all three process corners
~ `finishing` -> `signoff`
~ parasitics extraction and STA is own meta-step now instead of being lumped in routing, as part of the signoff block
~ Update OpenROAD
~ Update Open_PDKs, Sky130, and Magic
~ PDK verification made more PDK agnostic outside of env.py
~ Made RCX use LEF files instead of multiple configuration variables/set_rc.tcl
~ prep_lefs more or less rewritten
- Removed redundant RCX variables
- Removed `-order_wires` from various `read_def`s in OpenROAD- deprecated, apparently

[ci ets]

---
~~Blocked on https://github.com/RTimothyEdwards/open_pdks/pull/231/.~~
~~Blocked on https://github.com/RTimothyEdwards/open_pdks/issues/235.~~
~~Blocked on https://github.com/RTimothyEdwards/open_pdks/issues/239.~~

Resolves #1020


kareefardi (changes requested):
update sta.tcl and sta_multi_corner.tcl

designs/digital_pll_sky130_fd_sc_hd/config.tcl
```diff
@@ -8,7 +8,7 @@ set ::env(VERILOG_FILES) $::env(DESIGN_DIR)/src/digital_pll.v
 set ::env(SYNTH_READ_BLACKBOX_LIB) 1
 
 # Fill this
-set ::env(CLOCK_PERIOD) "15.8"
+set ::env(CLOCK_PERIOD) "18"
```
kareefardi, line 11: Are we free to mess with the clock period ?

flow.tcl
```diff
@@ -55,6 +55,10 @@ proc run_routing_step {args} {
 	run_routing
 }
 
+proc run_parasitics_sta_step {args} {
```
kareefardi, line 58: This function isn't used?

run_designs.py
```diff
@@ -249,7 +249,7 @@ def run_design(designs_queue):
                 "-overwrite",
                 "-no_save",
                 "-run_hooks",
-            ] + (["-verbose", "99"] if show_log_output else [])
+            ] + (["-verbose", "1"] if show_log_output else [])
```
kareefardi, line 252: Perhaps an future enhancement is exposing this at the cli level

donn: sure whatever

flow.tcl
donn: Oops. Will fix that.

designs/digital_pll_sky130_fd_sc_hd/config.tcl
donn: It's not passing otherwise.

scripts/tcl_commands/sta.tcl
```diff
@@ -48,4 +62,62 @@ proc run_sta {args} {
...
 
+proc run_parasitics_sta {args} {
+    set options {
+        {-sdf_out optional}
+    }
+    set flags {}
+    parse_key_args "parasitics_sta" args arg_values $options flags_map $flags
+
+    set_if_unset arg_values(-sdf_out) [file rootname $::env(CURRENT_DEF)].sdf
+    set_if_unset arg_values(-spef_out_prefix) [file rootname $::env(CURRENT_DEF)]
```
kareefardi, line 73: `spef_out_prefix` isn't declared in the `options` variable. I am unsure if passing this argument, when calling the function, will have any effect.

kareefardi (approved):

PR #991, opened 2022-03-09
configuration/README: fix TAKE_LAYOUT_SCROT default

proppy:
This seems off by default:
https://github.com/The-OpenROAD-Project/OpenLane/blob/e40929f867497967eb7048302bd434bcc598b4e1/configuration/general.tcl#L54


donn (approved):
Yep. Thank you for catching that.

PR #1006, opened 2022-03-16
cts force clk nets

kareefardi:
Add `CTS_FORCE_CLK_NETS` that overrides clock nets defined by the provided sdc + other small logging message changes.

The motivation behind this is that the sdc file sometimes doesn't properly define clock net. This leads to unwanted behavior during cts (and perhaps in other stages too like sta). Arguably, one should go fix the sdc file, but I see no harm to have this new variable as a workaround.


kareefardi:
Yes. I agree. I am having second thoughts about this and I changed it to draft. Especially, when I realized that it's likely I misunderstood CTS log file.

PR #1009, opened 2022-03-17
Fix -halo_width_y

kareefardi:
`-halo_width_y` was set `FP_TAP_HORIZONTAL_HALO` to instead of `FP_TAP_VERTICAL_HALO `


donn (approved):
Wow, that's silly. Thanks.

PR #1018, opened 2022-03-21
Add CTS_DISTANCE_BETWEEN_BUFFERS option

antonblanchard:


antonblanchard:
I'm struggling with CTS on a design where the clock tree delay ends up very large. It looks like OpenROAD is inserting too many buffers.

One way I've been able to improve things is to set `-distance_between_buffers`. Longer term CTS needs to be improved, but this at least offers a temporary solution.

maliberty:
@antonblanchard if you have an example of the problem it would be good to open an OR issue.  CTS definitely needs work

donn (approved):

PR #1024, opened 2022-03-29
Repair max slew and max capacitance violations at all resizer stages

antonblanchard:
Placement and global routing could create new max slew and max capacitance
violations, so call repair_design at all resizer stages.

---
[ci ets]


maliberty:
repair_design does not operate on clock nets so at least that much is ok.

maliberty:
@donn this would need CI to establish if it is beneficial in general.

donn:
@maliberty I'll run the full suite.

donn:
I'd actually say that's an improvement.

donn (approved):

PR #1027, opened 2022-03-30
tcl_commands: source install-wide config

proppy:
Not sure if that's the best way to achieve this, but I find myself willing to override some of the config parameters system-wide to correspond to a specific OpenLane deployment, e.g:
```
RUN_KLAYOUT=0
RUN_CVC=0
```

It seems that the settings in `install/env.tcl/ are not currently into account by the flow since https://github.com/The-OpenROAD-Project/OpenLane/blob/386d6042ae514b2bb4a5032ee121f9ac6e4aecca/scripts/tcl_commands/all.tcl#L316 will override them with their respective default value.

Let me know what you think!

### DIFF
--- a/scripts/tcl_commands/all.tcl
+++ b/scripts/tcl_commands/all.tcl
@@ -400,6 +400,11 @@ proc prep {args} {
 
     source $scl_config
 
+    # source install-wide config
+    if { [file exists $::env(OPENLANE_ROOT)/install/env.tcl ] } {
+        source $::env(OPENLANE_ROOT)/install/env.tcl
+    }
+
     # needs to be resourced to make sure it overrides the above
     source_config $::env(DESIGN_CONFIG)
 


proppy:
> That indeed is not the right place to do it, imo. I'm open to sourcing ~/.openlane.

I was looking for something system-wide (not per user) so that the configuration can match a given OpenLane install; `/etc/openlane/`? ;)

donn:
Not opposed to that, /etc/openlane/config.tcl and ~/.openlane/config.tcl sound like sensible additions. Can you do those?

proppy:
Maybe we could also make those configuration variables, so that they could be easily be overriden for non-standard installs.

donn:
`OPENLANE_GLOBAL_RC` and `OPENLANE_USER_RC`?

donn:
Any updates?

donn:
Staled out.

proppy:
Looks like just appending file to https://github.com/The-OpenROAD-Project/OpenLane/blob/master/configuration/load_order.txt should be enough to enable this.

PR #1031, opened 2022-04-04
sta: report_units on a run

kop316:
In working with other tools, units are reported on every run in the logs
so that the user can look back in the logs to understand units.

This allows OpenROAD to have a similar feature.


maliberty:
This allows **openlane** to have a similar feature  [Openroad is the app; openlane the flow using it]

kop316:
ok, fixed

PR #1033, opened 2022-04-04
Revert #1024

antonblanchard:
This reverts commit 2f276c62c30661f1ffd7b2c0cc721bb62ce2306e.
Amongst other things, repair_design undoes all the hold violation fixes
previously done.


antonblanchard:
@donn @maliberty I only just realised repair_design calls `Resizer::resizeToTargetSlew()` which calls `Resizer::findTargetCell()` which then considers all buffers equivalent (buf, clkbuf, delay buffers etc). So calling it later in the flow is bad, because it might convert delay buffers back to normal buffers (undoing hold violation fixups).

However I'm not sure why repair_design wants to substitute one type of buffer for another at all. It does try to avoid inserting delay buffers, but it only works sometimes (I see lots of delay buffers after a repair_design):

```
            // Library may have "delay" buffers/inverters that are
            // functionally buffers/inverters but have additional
            // intrinsic delay. Accept worse target load matching if
            // delay is reduced to avoid using them.
            ? ((delay < best_delay
                && dist < best_dist * 1.1)
               || (dist < best_dist
                   && delay < best_delay * 1.1))
            : dist < best_dist
```

maliberty:
@tspyrou FYI on rsz issue

donn (approved):

PR #1036, opened 2022-04-05
Run hold repair after setup repair

antonblanchard:
ORFS runs hold repair after setup repair, which makes sense considering they
are more critical.


antonblanchard:
@maliberty @jjcherry56 flipping the order makes sense to me but wanted to run it by you both.

jjcherry56:
repair_timing with no args does setup than hold but I would be surprised if it really mattered. Hold repair takes pains to avoid creating setup violations and it is hard to see how repairing setup on long paths is going to create hold issues.

donn (approved):

Issue #1037, opened 2022-04-09
OpenLane now fails on PDN with latest OpenROAD changes


cbalint13:
OpenLANE fails on PDN step due to OpenROAD [pdn: convert to c++ #1749](https://github.com/The-OpenROAD-Project/OpenROAD/pull/1749) .

Error [with a reproductible](https://github.com/The-OpenROAD-Project/OpenLane/files/8456577/pdn-issue.tar.gz) step:
```
$ openlane-flow.tcl -design spm
OpenLane 2022.03.31.02.35.19-20220401.0.git2f276c62.fc37
{...}
[STEP 6]
[INFO]: Generating PDN...
[ERROR]: during executing openroad script /usr/share/openlane/scripts/openroad/pdn.tcl
[ERROR]: Exit code: 1
[ERROR]: Last 10 lines:
[INFO ODB-0130]     Created 36 pins.
[INFO ODB-0131]     Created 478 components and 2698 component-terminals.
[INFO ODB-0133]     Created 352 nets and 984 connections.
[INFO ODB-0134] Finished DEF file: /home/cbalint/rpmbuild/SPECS/MAIN/RUN_2022.04.09_11.42.16/results/floorplan/spm.def
[WARNING PDN-1024] -verbose has been deprecated.
[WARNING PDN-1000] Using legacy PDNGEN.
Consider using "convert_pdn_config /usr/share/openlane/scripts/openroad/pdn_cfg.tcl" to convert the legacy configuration.
[ERROR PDN-0175] Pitch 28.8270 is too small for, must be atleast 28.8300
Error: pdn_cfg.tcl, 39 PDN-0175
{...}
```
OpenROAD version:
```
$ openroad -version
2.0-20220407.git34e0ea38.fc37 
```

---

All [openlane builds fails](https://copr.fedorainfracloud.org/coprs/rezso/VLSI/build/4123317) at post-build regression tests ```%check``` stage. 
Unsure what possible solution would be beyond just having ```convert_pdn_config```.

Cc @arlpetergadfort , @maliberty 

---

FYI, there is a little flow with the native .rpm repo usage: https://gist.github.com/cbalint13/1de29fbe77e8dedddd3861b5f49a1835

Thanks,
~cristian.

vijayank88:
@cbalint13 
Currently checked with OR version: OpenROAD v2.0-3078-g4e7b85576 
Not facing any issue with test case attached. Successfully completed PDN stage.

```
[...]

PR #1038, opened 2022-04-09
Switch to the new PDN from OpenROAD.

cbalint13:
This PR fixes issue #1037 related to ```openroad``` new PDN behaviour.

~~(1) By two decimal rounding it fixes the error:~~
```
[ERROR PDN-0175] Pitch 28.8270 is too small for, must be atleast 28.8300
```

(2) Final conversion to new PDN part fixes the error:
```
Consider using "convert_pdn_config /home/cbalint/{...}/pdn_cfg.tcl" to convert the legacy configuration.
[INFO PDN-9008] Design name is spm.
[INFO PDN-9009] Reading technology data.
[ERROR PDN-9017] No stdcell grid specification found - no rails can be inserted.
```
---

This now pass latest openlane [builds & regression](https://copr.fedorainfracloud.org/coprs/rezso/VLSI/build/4144274/) .
~~Q: do we need discriminate within ```pdn.tcl``` the **old** vs **new** PDN or just lets go with it forward (drop old) ?~~

Cc @arlpetergadfort , @maliberty

Thanks,
~cristian.

### DIFF
--- a/scripts/openroad/pdn.tcl
+++ b/scripts/openroad/pdn.tcl
@@ -32,7 +32,14 @@ if {[catch {read_def $::env(CURRENT_DEF)} errmsg]} {
     exit 1
 }
 
-if {[catch {pdngen $::env(PDN_CFG) -verbose} errmsg]} {
+# load the grid definitions
+if {[catch {source $::env(PDN_CFG)} errmsg]} {
+    puts stderr $errmsg
+    exit 1
+}
+
+# run PDNGEN
+if {[catch {pdngen} errmsg]} {
     puts stderr $errmsg
     exit 1
 }


scripts/openroad/pdn.tcl
```diff
@@ -32,7 +32,12 @@ if {[catch {read_def $::env(CURRENT_DEF)} errmsg]} {
     exit 1
 }
 
-if {[catch {pdngen $::env(PDN_CFG) -verbose} errmsg]} {
+if {[catch {convert_pdn_config $::env(PDN_CFG)} errmsg]} {
+    puts stderr $errmsg
+    exit 1
+}
```
arlpetergadfort, line 38: This command doesn't actually do anything. It's provided as a convenience to make the switch from the old to the new. The correct thing to do there is to source the `$::env(PDN_CFG)`.

scripts/openroad/pdn.tcl
```diff
@@ -32,7 +32,12 @@ if {[catch {read_def $::env(CURRENT_DEF)} errmsg]} {
...
+
+if {[catch {pdngen -verbose} errmsg]} {
```
arlpetergadfort, line 40: `-verbose` has been removed so you will just keep getting warnings about that if it's kept.

scripts/tcl_commands/floorplan.tcl
```diff
@@ -69,11 +69,11 @@ proc init_floorplan {args} {
 			$core_height <= [expr {$::env(FP_PDN_HOFFSET) + $::env(FP_PDN_HPITCH)}]} {
 				puts_warn "Current core area is too small for a power grid. The power grid will be minimized."
 
-			set ::env(FP_PDN_VOFFSET) [expr {$core_width/6.0}]
-			set ::env(FP_PDN_HOFFSET) [expr {$core_height/6.0}]
+			set ::env(FP_PDN_VOFFSET) [expr {double(round(100*($core_width)/6.0))/100}]
+			set ::env(FP_PDN_HOFFSET) [expr {double(round(100*($core_height)/6.0))/100}]
```
arlpetergadfort, line 73: I would probably be better to look at the manufacturing grid and use that in case it's something different.

scripts/openroad/pdn.tcl
cbalint13: Right, I updated the change to just ```source```.

scripts/openroad/pdn.tcl
cbalint13: Removed.

scripts/tcl_commands/floorplan.tcl
cbalint13: * My feeling was that is simply a rounding error from very 28.826666... to improper (```28.8270``` / 3 dec) instead of (```28.8300``` / 2 dec), so rounding to expected 2 dec now works.

Let me know if still unacceptable.

cbalint13: @arlpetergadfort ,

Double checked:

* Another message: ```Pitch 30.5130 is too small for, must be atleast 30.5200```
* Looks like really is a mismatch in expectation of 2-digits vs 3-digits.
* I simply forced in .tcl a rounding to two digit (and does the trick), no other magic is in the patch.

cbalint13: @arlpetergadfort 

* updated to a more elegant way of formatting to 2 decimal.

Let me know if still unacceptable.

maliberty: It doesn't make sense to something coarser than the manufacturing grid otherwise you would disallow legal inputs.

maliberty: Personally I would prefer it be an error rather than rounding as the user is clearly doing something wrong with an off-grid value.  Guessing how to fix is dangerous.

cbalint13: * Removed (1) about the "rounding of division" that lead to part, lets get only PDN related behavior fixed for now.

cbalint13:
@maliberty , @arlpetergadfort 

Let's leave out from here the very [SPM](https://github.com/The-OpenROAD-Project/OpenLane/tree/master/designs/spm),  will need a separate fix (another PR) for the grid / rounding issue.

---

Q: For my understanding (for future fix), for the case of [spm design sample](https://github.com/The-OpenROAD-Project/OpenLane/tree/master/designs/spm), having ```FP_PDN_AUTO_ADJUST``` :

   The grid i.e. 0.005 must be perfect divisor of  that ```Pitch``` with/height numbers ?

  * If so, actual div() by 3.0 (for ```spm``` sample) in [floorplan.tcl](https://github.com/The-OpenROAD-Project/OpenLane/blob/f2f5d8358795a14f3eb1c7c9b0a992b2ab2c49f0/scripts/tcl_commands/floorplan.tcl#L72-L76) leads to infinite decimal , inevitably will round somewhere.
  * The final roundings of the div() (e.g 28.826666...) will never divide/align a fixed grid of 0.00X a fixed decimal number.
 
 Thanks,
~cristian.

maliberty:
@donn I do not recommend merging this.  @arlpetergadfort is working on making off-grid into an error.   I do not think pdn should snap the user input.

I don't know why you are dividing by 3 but you should snap the result to the grid.

cbalint13:
> @donn I do not recommend merging this. @arlpetergadfort is working on making off-grid into an error. I do not think pdn should snap the user input.
> 
> I don't know why you are dividing by 3 but you should snap the result to the grid.

@donn

Wait, @maliberty:

   1. I dropped out from here the part with roundings (will see another PR for that) !
   2. Let'ts just merge remaining part "source $::env(PDN_CFG) + pdngen" as **new behavior**, proposed by @arlpetergadfort .
   
Please have another quick  look on this PR final diff summary.

maliberty (approved):

scripts/tcl_commands/floorplan.tcl
proppy: @tdene and I were wondering if you will open another PR with the fix for `floorplan.tcl`: https://github.com/The-OpenROAD-Project/OpenLane/pull/1038/files/5b1d54789d3a17c66ba73964e9d7961d2decfacb#diff-3b1e9d4927839bb0fe38be8ece41a6d968939bc388a109515846167b96c9ad4d

PR #1039, opened 2022-04-11
Global Placement Fixes

antonblanchard:


scripts/openroad/replace.tcl
```diff
@@ -66,15 +66,17 @@ if { $::env(PL_BASIC_PLACEMENT) } {
 }
 
 if { $::env(PL_TIME_DRIVEN) } {
+	source $::env(SCRIPTS_DIR)/openroad/set_rc.tcl
 	read_sdc $::env(CURRENT_SDC)
 	read_verilog $::env(synthesis_results)/$::env(DESIGN_NAME).v
```
maliberty, line 71: While it isn't part of your change I need to ask why is the verilog being read here after the DEF has already been loaded?  @donn

antonblanchard: On that topic - I wonder if it would be better to load everything at the start unconditionally (perhaps with a helper script), vs loading things as we need them, based on config variables. As an example, we did call set_rc.tcl, just later on in the script.

antonblanchard: https://github.com/The-OpenROAD-Project/OpenLane/pull/1041 is a first step to be more consistent in how we load files. tech liberty, other liberty files, lef, def.

donn: @maliberty Pre-dates me. My git blame is just me updating an environment variable.

maliberty: @donn can you remove the read_verilog and see if that has any effect?

antonblanchard: I removed a few unncessary calls to `read_verilog` in https://github.com/The-OpenROAD-Project/OpenLane/pull/1051

donn: Removed.

maliberty (approved):

antonblanchard:
Hi @donn, any chance we could merge this? I've been experimenting with enabling `PL_ROUTABILITY_DRIVEN` (similar to ORFS) and it needs this patch to do reasonable things.

donn:
@antonblanchard I asked a question in my review.

antonblanchard:
> @antonblanchard I asked a question in my review.

I'm might be missing something obvious, but I can't see the question above.

donn:
2D47E7E6-4D8A-4025-8460-817700B2F3AF

antonblanchard:
@donn I don't see that comment at all. It's marked `Pending` on your image, that might have something to do with it.

donn:
Hmm. Can I blame GitHub's UI for that one? :upside_down_face:

antonblanchard:
As to why I added a call to `set_rc`, the short answer is we do it everywhere else so I wanted to be consistent.

For the long answer... I've been trying to understand if the TLEF capacitance data is adequate for pre SPEF parasitic capacitance estimates and I'm not sure it is. The TLEF spec has figures for plate capacitance and edge capacitance, but what does that edge capacitance even represent? I presume it is a mix of fringe and coupling capacitance, but that is going to vary a lot depending on the density of the design and how close other wires are.

In ORFS there are a set of scripts to take SPEF data from routed designs, and back calculate the layer capacitance estimates. This should provide a better estimate than the TLEF data I think. This process also allows a design that is very different (eg extremely dense, or extremely sparse) to do a first pass to calculate better layer capacitance values, and use them in subsequent tape outs. A lot of guesswork on my side, so @jjcherry56 @maliberty feel free to correct me :)

jjcherry56:
I would say an even bigger motivation for correlating the global routing rc estimations to spef values with set_layer_rc is to guarentee consistency with rcx (spef) which is in turn correlated with commercial extraction. The TLEF values do not take context into account (wires above, below and adjacent). The correlation matches a layer rc per micron that includes the context for all the wires in the ORFS tests. The other issue that was a constant headache was determining which corner/min/max values in the datasheet correspond to the pdk commercial extraction values.

donn:
At any rate, sky130 no longer includes values for `set_rc`. So I'll just go ahead and merge this- the reasoning "for consistency" is good enough for me.

maliberty:
> At any rate, sky130 no longer includes values for `set_rc`. So I'll just go ahead and merge this- the reasoning "for consistency" is good enough for me.

@donn why does it no longer include set_rc values?

PR #1044, opened 2022-04-12
Avoid off-grid errors in FP_PDN_AUTO_ADJUST

maliberty:
Change /3 -> /4 and /6 -> /8 to help avoid offgrid errors.  Really
whatever is used should be snapped to the manufacturing grid. This should
be done inside openroad where the LEF data is available.  This is just
less likely to cause an error.

### DIFF
--- a/scripts/openroad/pdn_cfg.tcl
+++ b/scripts/openroad/pdn_cfg.tcl
@@ -24,7 +24,7 @@ if { [info exists ::env(FP_PDN_ENABLE_GLOBAL_CONNECTIONS)] } {
 
 set_voltage_domain -name CORE -power $::env(VDD_NET) -ground $::env(GND_NET)
 
-# Assesses whether the deisgn is the core of the chip or not based on the 
+# Assesses whether the design is the core of the chip or not based on the 
 # value of $::env(DESIGN_IS_CORE) and uses the appropriate stdcell section
 if { $::env(DESIGN_IS_CORE) == 1 } {
     # Used if the design is the core of the chip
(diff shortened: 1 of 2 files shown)


proppy:
This does solve #1037 for me, see https://colab.research.google.com/gist/proppy/fce76d98e4f21d4c1a9db08d2d991e94/openlane-with-conda-eda.ipynb

proppy:
@maliberty do we already have an issue on the OpenROAD side to capture this?

maliberty:
This is more of a quick patch than a proper fix.

tdene:
@maliberty I started writing a PR last night, before you submitted yours.

My thoughts were the following:
- Expose the manufacturing grid size in some fashion. The only place it's currently used is in [scripts/defutil.py](https://github.com/The-OpenROAD-Project/OpenLane/blob/7891c2c8859e8c800226920c181fa039ad298e79/scripts/defutil.py), so what I had started doing was just writing as quick `get_manufacturing_grid` method there.
 - When the time comes, in floorplan.tcl, to check for `FP_PDN_AUTO_ADJUST` and adjust accordingly, first call the script that gets the manufacturing grid size.
 - Then snap the relevant variables to the grid.

Does this sound reasonable?

If so, this comment could be used as a future recipe for any parties interested in implementing a full fix to this issue.

proppy:
> @proppy this isn't an OR issue but OL generating bad input to OR.

ah my bad, I think miss-interpreted your earlier comment "This (snapping to manufacturing grid) should be done inside openroad where the LEF data is available": did you meant that OpenROAD should expose enough information so that OpenLane can perform the snapping?

maliberty:
@proppy 
> Did you mean that OpenROAD should expose enough information so that OpenLane can perform the snapping?

Yes, in the OL tcl you don't have the grid available to do the snapping

maliberty:
@tdene that would work.  I was thinking it might be easier to just move that section of code into openroad/pdn.tcl where you are already inside OR and have access to the grid.   I don't think any other code cares about the values.

donn (dismissed):

donn (approved):

PR #1046, opened 2022-04-12
Update PDN Configurations

maliberty:
Update pdn_cfg.tcl to use only commands supported by the new pdngen

The prior code was a mix of new commands and old tcl vars that are no
longer supported by the new pdngen.  I find the old setup confusing
so this is my best guess at a correct replacement.

---
Notes from maintainer:

Resolves #1037.

Will be merged after mpw-6 is tagged so we don't break all the caravel scripts.

PR #1059, opened 2022-04-16
Update to use the new C++ pdngen in OR

maliberty:
designs/inverter: updated die area to be a multiple of site size
designs/s44: has no need for a custom cfg so just removed it


maliberty:
Recreates #1046 which was mistakenly closed/removed.

donn (approved):
Once again, approved, but will merge after the final mpw-6 tag.

cbalint13:
> Once again, approved, but will merge after the final mpw-6 tag.

@maliberty , @donn 

* The PR fails on ```check_power_grid``` with sample ```spm``` design:
* Attached the [reprod.tar.gz](https://github.com/The-OpenROAD-Project/OpenLane/files/8505727/reprod.tar.gz) .
* PR was applied to ```85b6ceca``` (current latest head).

```
$ openlane-flow.tcl -design spm
{...}
[STEP 6]
[INFO]: Generating PDN...
[ERROR]: during executing openroad script /usr/share/openlane/scripts/openroad/pdn.tcl
[ERROR]: Exit code: 1
[ERROR]: Last 10 lines:
[WARNING PSM-0030] VSRC location at (43.240um, 42.160um) and size 10.000um, is not located on an 
existing power stripe node. Moving to closest node at (48.760um, 31.960um).
[INFO PSM-0031] Number of PDN nodes on net VGND = 450.
[INFO PSM-0064] Number of voltage sources = 1.
[WARNING PSM-0038] Unconnected PDN node on net VGND at location (5.280um, 95.200um), layer: 6.
[WARNING PSM-0038] Unconnected PDN node on net VGND at location (92.240um, 95.200um), layer: 6.
[WARNING PSM-0038] Unconnected PDN node on net VGND at location (92.000um, 10.640um), layer: 5.
[WARNING PSM-0038] Unconnected PDN node on net VGND at location (92.000um, 95.440um), layer: 5.
[ERROR PSM-0069] Check connectivity failed.
Error: pdn.tcl, 51 PSM-0069
child process exited abnormally
```
```
$ openroad -version
2.0-20220417.git4a67ed48.fc37 
```
```
$ rpm -qa | grep openlane
openlane-2022.04.17.01.59.58-20220416.0.git85b6ceca.fc37.noarch
openlane-designs-2022.04.17.01.59.58-20220416.0.git85b6ceca.fc37.noarch
```

maliberty:
@cbalint13 this PR shows spm passing in the CI.  How is your run different?

cbalint13:
> @cbalint13 this PR shows spm passing in the CI. How is your run different?
@maliberty , @donn 

* Looked at CI results, unclear for me what openroad it use ( I guess the one from [tool_metadata.yml](https://github.com/The-OpenROAD-Project/OpenLane/blob/master/dependencies/tool_metadata.yml) ?! ).
* I am getting the error with latest HEAD of openroad (overriding/ignore any ```tool_metadata.yml``` checks)

maliberty:
@cbalint13 what commit of OR did you use?  The head keeps moving along.

maliberty:
The offset is off grid so my /4 hack doesn't work well for spm - probably because it is so small.

[INFO]: Final Vertical PDN Offset: 11.4425

kareefardi:
@maliberty I noticed that you got rid of `FP_PDN_ENABLE_MACROS_GRID` and `FP_PDN_MACROS`. The former is just a flag enabling and disabling macro power grid while the latter is a result of `FP_PDN_MACRO_HOOKS`.

maliberty:
> @maliberty I noticed that you got rid of `FP_PDN_ENABLE_MACROS_GRID` and `FP_PDN_MACROS`. The former is just a flag enabling and disabling macro power grid while the latter is a result of `FP_PDN_MACRO_HOOKS`.

I don't see the purpose of FP_PDN_ENABLE_MACROS_GRID.  It doesn't actually turn off the grid as it has an else that still defines a grid.  I'm glad to put it back if you can explain what it really does.

In general there is no benefit in defining a separate grid per macro when they have the same settings (in fact it will run slower).

maliberty:
> Another thing is that the section below sets the default value for `VDD_NET` to the value of `VDD_PIN` and similarly it did the same with `GND`. I am not saying that's correct. I am just highlighting a difference that I noticed.
> 
> ```
> if { ! [info exists ::env(VDD_NET)] } {
> 	set ::env(VDD_NET) $::env(VDD_PIN)
> }
> 
> if { ! [info exists ::env(GND_NET)] } {
> 	set ::env(GND_NET) $::env(GND_PIN)
> }
> ```

This are obsolete and handled by -net $::env(VDD_NET) (likewise gnd)

kareefardi:
> I don't see the purpose of FP_PDN_ENABLE_MACROS_GRID. It doesn't actually turn off the grid as it has an else that still defines a grid. I'm glad to put it back if you can explain what it really does.
> 
> In general there is no benefit in defining a separate grid per macro when they have the same settings (in fact it will run slower).

I agree that `FP_PDN_ENABLE_MACROS_GRID` seems a bit not needed. 

On the other hand, it is still possible to have a top level macro with multiple power domains where separate macro are driven by separate domains and that is what I believe `FP_PDN_MACRO_HOOKS` allowed and could be achieved using the new pdngen.


> This are obsolete and handled by -net $::env(VDD_NET) (likewise gnd)

I mean that `VDD_NET` and `GND_NET` in this PR currently don't have default values and designs don't initialize these values which makes me wonder whether `add_global_connections` is being called at all.

maliberty:
> 



> > I don't see the purpose of FP_PDN_ENABLE_MACROS_GRID. It doesn't actually turn off the grid as it has an else that still defines a grid. I'm glad to put it back if you can explain what it really does.
> > In general there is no benefit in defining a separate grid per macro when they have the same settings (in fact it will run slower).
> 
> I agree that `FP_PDN_ENABLE_MACROS_GRID` seems a bit not needed.
> 
> On the other hand, it is still possible to have a top level macro with multiple power domains where separate macro are driven by separate domains and that is what I believe `FP_PDN_MACRO_HOOKS` allowed and could be achieved using the new pdngen.

That would require the generation of multiple domains which I didn't see the previous code doing.  Is there any test of this functionality?  I'm suspicious about whether it works as described.

> 
> > This are obsolete and handled by -net $::env(VDD_NET) (likewise gnd)
> 
> I mean that `VDD_NET` and `GND_NET` in this PR currently don't have default values and designs don't initialize these values which makes me wonder whether `add_global_connections` is being called at all.

You are right, I removed a bit too much here , I only meant to remove "set ::power_nets $::env(VDD_NET)".  I've restore the default setting of VDD_NET (likewise gnd).

maliberty:
@donn I think inverter is real but everything else looks like false magic violations.  What do you suggest I do?

donn:
@maliberty Send a reproducible to @RTimothyEdwards

maliberty:
@donn @kareefardi the tests are now passing, please review and merge if possible.

donn:
@kareefardi All yours buddy.

cbalint13:
@donn, @kareefardi , @maliberty

* Confirming that this PR  +openroad HEAD fixes issue: [#1037](https://github.com/The-OpenROAD-Project/OpenLane/issues/1037) .
* One introduced workaround isn't even necessary PR: [#1044](https://github.com/The-OpenROAD-Project/OpenLane/pull/1044) (does not matter).

Sample build + test logs: [here](https://download.copr.fedorainfracloud.org/results/rezso/VLSI/fedora-rawhide-x86_64/04369988-openlane/builder-live.log.gz) .

maliberty:
@antonblanchard I've updated the OR commit id to get the dpl fix.

kareefardi (approved):

kareefardi:
Thanks @maliberty for incorporating macro hooks variable. This looks fine.
I am wondering what's the use for `-secondary` in power domains and why wouldn't we define different power domains and different grids. We can continue this discussion somewhere else if you prefer.

maliberty:
@kareefardi there is just one core domain.  Defining multiple domains can only be done in multiple runs as you were doing before which is inefficient.  Multiple domains are intended for macros, power switching, etc where there is a locally separate grid structure.

kareefardi:
@maliberty Yes I realized that after posting the comment. Is it even possible to define multiple domains in multiple runs? I recall getting this error  `Pin vccd1 of instance mprj2_logic_high_inst is not connected to any net.` Is that due to having a domain completely disconnected in one of the multiple runs?

PR #1061, opened 2022-04-18
No need to check branch name in get_tag.py

antonblanchard:
get_tag.py currently fails if you explicitly check out a commit (it
returns "-dev").

Now we are using commit based tags, there is no need to check for the
current branch.

### DIFF
--- a/dependencies/get_tag.py
+++ b/dependencies/get_tag.py
@@ -37,21 +37,6 @@ def get_tag() -> str:
                 "Cannot find the git binary. Please specify OPENLANE_IMAGE_NAME manually."
             )
 
-        branch_name_data: subprocess.CompletedProcess = subprocess.run(
-            ["git", "branch", "--show-current"],
-            stdout=subprocess.PIPE,
-            stderr=subprocess.PIPE,
-        )
-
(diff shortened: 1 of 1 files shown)


donn:
I don't agree. It keeps development branches named explicitly as such. What is the motivation for this PR?

antonblanchard:
Sounds like I'm doing something wrong. I'm trying to work out how I can pin a particular version of Openlane, so future builds of a tape out use the same toolchain. Neither of these work:

```
# git checkout 85b6ceca9e38a2187226457ad398f8da3353e8e5
# dependencies/get_tag.py
-dev

# git checkout 2022.04.17_01.59.58
# dependencies/get_tag.py
-dev
```

donn:
Well, in this situation we'd typically use OPENLANE_IMAGE_NAME: but maybe get_tag.py should attempt to detect a detached HEAD. I'll look at it.

antonblanchard:
I can use `OPENLANE_IMAGE_NAME`. If nothing else, it might be nice for this to error out - it took me a while to realise `make` was not installing the PDK but rebuilding the entire docker image :)

donn:
Implemented in #1062.

PR #1073, opened 2022-04-27
Design configuration file tweeks

marwaneltoukhy:
Added an extra layer of protection to make sure that the configurations are loaded correctly from the PDK library specific config files.


donn:
No. That's not the intent. If a file is not found, then we should just assume that there's no specific configuration for this PDK.

PR #1087, opened 2022-05-11
Various Routing Documentation Updates

proppy:
/cc @antonblanchard


maliberty:
What's the motivation for this change?

proppy:
sorry for the not being explicit in the change description, this simply correct the documentation to match the default TCL configuration.

PR #1091, opened 2022-05-13
Update FP_PDN_MACRO_HOOKS documentation

antonblanchard:
There is no default for FP_PDN_MACRO_HOOKS, so remove the comment suggesting
there is.


donn:
Merged in #1091

PR #1092, opened 2022-05-13
Enable PL_ROUTABILITY_DRIVEN and PL_TIME_DRIVEN

antonblanchard:
ORFS enables both of these by default.


maliberty (approved):
We are supporting it

PR #1098, opened 2022-05-17
Remove DONT_BUFFER_PORTS

### LINKED ISSUE #1095: Add OpenROAD remove_buffers to flow
@jjcherry56 points out that we should have the option of calling `remove_buffers` before we call `repair_design`, to undo any buffer addition that yosys/abc does.

To complicate matters, Openlane currently has an internal command called `remove_buffers`. It's controlled by an undocumented environment variable `DONT_BUFFER_PORTS`. What is it used for? @donn any ideas?

I have a distant memory that it was to remove buffers inserted around tristates. I did find it a mention of it in caravel, but it's commented out:

```
# mgmt_gpio_in is driven by a tristate cell
# set ::env(DONT_BUFFER_PORTS) "mgmt_gpio_in"
```

If we are inserting buffers around tristates, I would think we should fix OpenROAD, and just remove this.

antonblanchard:
DONT_BUFFER_PORTS was used to work around a bug in OpenROAD where it would
add buffers to tristate nets. Since this issue is fixed (and the workaround
is no longer used in caravel), remove it.

---
Resolves #1095


kareefardi:
I am just highlighting that one extra things offered by DONT_BUFFER_PORTS is that you have control over which output ports to remove buffers from .

donn:
@kareefardi I'll leave reviewing this one to you since it's more your domain.

antonblanchard:
@kareefardi if it is still useful, perhaps we can something to the documentation.

antonblanchard:
@jjcherry56 makes sense. @kareefardi do you have a way to test this feature? Could you make the suggested change and test if it still works?

donn:
Sorry, we can't merge this.

This feature is needed for some internal designs where the resizer would add buffers to output nets. We can't find a method to declare a net as untouchable. If there is one such method in OpenROAD, @maliberty, please point us to it.

I will rename the fn explicitly to `remove_port_buffers` and clearly label it as a workaround for something.

PR #1105, opened 2022-05-23
Synthesis Parameters, More Report Metrics

growly:
1) Defining `$::env(SYNTH_PARAMETERS)` passes top-level verilog module parameters to the yosys synth script, so you can override them per run.
2) Core area, various post-parasitics power numbers and critical path are all extracted for the automatic reports `generate_reports.py` creates. I wonder if these should be hidden behind an option like `MORE_METRICS`?


donn (approved):
Looks OK

donn (changes requested):
Wait, wait. Can you please document the new option you've added?

donn:
I meant the `SYNTH_PARAMETERS` variable, but that I forgot about that file, so thank you for that. I went ahead and added SYNTH_PARAMETERS myself.

scripts/report/report.py
```diff
@@ -317,6 +331,84 @@ def re_get_last_capture(rx, string):
...
+                    r"^Total\s+([\d.Ee\-+]+)\s+([\d.Ee\-+]+)\s+([\d.Ee\-+]+)\s+([\d.Ee\-+]+).*$",
+                    line,
+                )
+                if match:
+                    power_multi_corner_sta[current_corner].update(
+                        {
+                            "internal": float(match[1]),
+                            "switching": float(match[2]),
+                            "leakage": float(match[3]),
+                            "total": float(match[4]),
```
proppy, line 378: was that deliberate to not add this to the csv?

/cc @growly

PR #1113, opened 2022-05-26
`repair_timing` improvements

antonblanchard:
~ Don't catch errors from repair_timing. There are many reasons it might
  fail and reporting "utilization limit is reached" is just confusing.
~ Handle arguments with an arg_list
Suggested by @jjcherry56


antonblanchard:
These were a couple of suggestions from @jjcherry56

donn (approved):

PR #1133, opened 2022-06-07
Fix rdb drc output

antonblanchard:


donn (approved):
Looks okay, thanks anton

PR #1156, opened 2022-06-21
Fix ECO flow

antonblanchard:
The ECO flow has been broken for a number of months, so I thought I'd take a look.

I was pretty confused by where files ended up during the ECO flow, because the first iteration was different to the subsequent ones. This reworks it so that all iterations are directly out of `run_eco_flow`, and the ECO directories are  `X/eco_Y` (eg `logs/routing/eco_0`). When complete we copy back the output files from the last iteration back.

I've also made a decent performance improvement to the eco.py script (60s to 5s on my test).


donn (approved):
Looks great. Yeah, the ECO flow is horribly under-maintained.

PR #1166, opened 2022-06-26
io_placer.py fix for ZeroDivisionError

marwaneltoukhy:
If there's a side that doesn't have any ios the script breaks because of ZeroDivisionError


donn:
You literally just could've reviewed #1164

PR #1169, opened 2022-06-28
Update OpenROAD + Cleanup Fallout 

cbalint13:
+ new `openlane-and-push-tools` target added to root Makefile to enable easier updating of tools
~ antenna check results now extracted from log
~ `droute.tcl` updates to read guides before `detailed_route`

Original PR body follows:

---

This PR updates to the new ```check_antennas``` behaviour.

---
- Beggining with [OR#1979](https://github.com/The-OpenROAD-Project/OpenROAD/pull/1979) ```-report_file``` is now gone, see [ant docs](https://github.com/The-OpenROAD-Project/OpenROAD/blob/master/src/ant/README.md)
- Report is now parsable from the ```antenna.log``` instead of ```antenna.rpt```

Thanks,
~Cristian.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -90,6 +90,10 @@ all: get-openlane pdk
 openlane: venv/created
 	@PYTHON_BIN=$(PWD)/venv/bin/$(PYTHON_BIN) $(MAKE) -C docker openlane
 
+.PHONY: openlane-and-push
+openlane-and-push: venv/created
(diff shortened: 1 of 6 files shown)


donn:
The `caravel_upw` design is failing. Can you check what's up with that?

cbalint13:
@donn

Double trouble:

- The PR **fixes antenna checker** using the the latest openroad ```master```.
- But dragging in the latest openroad unveils a new different bug (not antenna-checker related).

---

For reference I expose here the **other issue**:

https://github.com/The-OpenROAD-Project/OpenLane/runs/7160808690?check_suite_focus=true#step:9:127
```Guide in net io_in[25] uses layer met5 (12) that is outside the allowed routing range [met1 (4), (met4)].```

PR #1172, opened 2022-06-29
Fix global placement cell padding

antonblanchard:
We were using the detailed placement interface.

### DIFF
--- a/scripts/openroad/replace.tcl
+++ b/scripts/openroad/replace.tcl
@@ -57,6 +57,7 @@ set arg_list [list]
 
 lappend arg_list -verbose_level 1
 lappend arg_list -density $::env(PL_TARGET_DENSITY)
+lappend arg_list -pad_right $::env(CELL_PAD)
 
 if { $::env(PL_BASIC_PLACEMENT) } {
 	lappend arg_list -overflow 0.9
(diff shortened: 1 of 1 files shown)


antonblanchard:
This is going to require many designs to update `PL_TARGET_DENSITY`. I'm not sure if there is a way to avoid it though.

donn:
What would be the damage on waiting on this PR for a bit? It looks like it's gonna need a lot more work and I get to go on vacation soonish

antonblanchard:
I think it can wait, since it is going to break quite a few things.

vijayank88:
@donn 
Can you please check this PR? 
Its need to be merged ASAP.

donn:
@vijayank88 
> Can you please check this PR?
> Its need to be merged ASAP.

Given that it breaks half the designs and I still have not had it adequately explain to me what the urgency is I wouldn't hold my breath.

antonblanchard:
We are applying zero padding during global placement, regardless of the setting. The fact detailed placement works at all with a non zero padding is just luck.

donn:
When did `set_placement_padding -global -right $::env(CELL_PAD)` stop working?

antonblanchard:
Global placement padding seems to have been added back in April 2020 via the `-pad_left` and `-pad_right` options:

https://github.com/The-OpenROAD-Project/RePlAce/commit/4c3e2954a4554b3ae110793067ca4e94b3de7860

Seems a lot like OpenROAD has been broken since day 0.

antonblanchard:
> When did `set_placement_padding -global -right $::env(CELL_PAD)` stop working?

`-global` in this context means all instances in detailed placement, not global placement doesn't it?

donn:
> -global in this context means all instances in detailed placement, not global placement doesn't it?

That's amazing.

donn:
Continued in #1226…

PR #1175, opened 2022-06-29
Reimplement support for multiple power domains

marwaneltoukhy:
`FP_PDN_MACRO_HOOKS` now takes the power and ground pin names in addition to instance name, power and ground nets.

This fix will enable pdn to connect to multiple power domains to internal macros with different power and ground pin names

### DIFF
--- a/configuration/README.md
+++ b/configuration/README.md
@@ -72,7 +72,7 @@ These variables are optional that can be specified in the design configuration f
 | `FP_PDN_CORE_RING` | Enables adding a core ring around the design. More details on the control variables in the pdk configurations documentation. 0=Disable 1=Enable. <br> (Default: `0`) |
 | `FP_PDN_ENABLE_RAILS` | Enables the creation of rails in the power grid. 0=Disable 1=Enable. <br> (Default: `1`) |
 | `FP_PDN_ENABLE_MACROS_GRID` | Enables the connection of macros to the top level power grid. 0=Disable 1=Enable. <br> (Default: `1`) |
-| `FP_PDN_MACRO_HOOKS` | Specifies explicit power connections of internal macros to the top level power grid. Comma separated list of macro instance names and power domain vdd and ground net names: `<instance_name> <vdd_net> <gnd_net>` |
+| `FP_PDN_MACRO_HOOKS` | Specifies explicit power connections of internal macros to the top level power grid. Comma separated list of macro instance names, power domain vdd and ground net names, and macro vdd and ground power pin names: `<instance_name> <vdd_net> <gnd_net> <vdd_pin> <gnd_pin>` |
 | `FP_PDN_CHECK_NODES` | Enables checking for unconnected nodes in the power grid. 0=Disable 1=Enable. <br> (Default: `1`) |
 | `FP_TAP_HORIZONTAL_HALO` | Specify the horizontal halo size around macros during tap insertion. The value provided is in microns. <br> Default: `10` |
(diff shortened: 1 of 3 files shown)


d-m-bailey:
@marwaneltoukhy `tcl_commands/floorplan.tcl` also uses `FP_PDN_MACRO_HOOKS`. Does it need to be changed too?

marwaneltoukhy:
@d-m-bailey No actually we don't need to change that, this part of the `tcl_commands/floorplan.tcl` just makes sure that specified net names are in the design.

vijayank88:
@marwaneltoukhy 
Till MPW6, all shuttle design maintained here:https://github.com/The-OpenROAD-Project/OpenLane-MPW-CI
Regression run against each OpenLane release for all shuttle designs.

With this commit I am facing error:
```
[ERROR]: FP_PDN_MACRO_HOOKS missing power and ground pin names
Error: pdn_cfg.tcl, 52 invoked "return" outside of a proc.
child process exited abnormally
```
Check CI logs here: https://github.com/The-OpenROAD-Project/OpenLane-MPW-CI/pull/58/checks?check_run_id=7171780891

Is there any variable to skip this error?

vijayank88:
```
       set instance_name [lindex $pdn_hook 0]
        set power_net [lindex $pdn_hook 1]
        set ground_net [lindex $pdn_hook 2]
        set power_pin [lindex $pdn_hook 3]
        set ground_pin [lindex $pdn_hook 4]
    ```
    Needs backward compatibility here. Because Till MPW6 PDN_HOOKS format as follows:
    ```
## Internal Macros
#### Macro PDN Connections
set ::env(FP_PDN_MACRO_HOOKS) "\
        user_proj_example_1 vccd1 vssd1"
```

kareefardi:
@vijayank88 In my opinion, exiting raises the user's awareness by requiring the user to check the power pins of the design's macros. We can provide a variable to skip the error and use the same names for power and ground pins (old behavior) but then if you are going to add that variable to the CI, why not fully define the macro hooks?

vijayank88:
@kareefardi 
For upcoming MPW shuttle user will follow that. 
Currently more than 30+ user repo had such configuration. So its difficult to modify it.
Earlier `set ::env(FP_IO_UNMATCHED_ERROR) 1` introduced to user must define all pins for placement. To skip `set ::env(FP_IO_UNMATCHED_ERROR) 0` given for backward compatibility.

So expecting such kind of configuration to support back-ward compatibility. Thanks

maliberty:
I already pointed out that you could make it accept either syntax by looking at the length.

marwaneltoukhy:
This is intentional and it is done to make users aware of what they are doing, and to force them to double check their designs and specifically set the power pins. If we add a variable to skip this check all users will use it and we'll get all kinds of LVS errors thrown at us because of users skipping that check. The error is clear enough and anyone trying to reharden a project from an old shuttle will have to do some modifications in their openlane configuration anyways, so I don't see a problem with it.

donn:
I agree with @marwaneltoukhy. This is an acceptable regression.

vijayank88:
Yes I agree the changes @marwaneltoukhy. Those who are going to submit MPW7 and re-submit design will follow the latest changes. But how to run tape-out passed design in CI. We can't reach out each user to raise PR and update the required changes.

vijayank88:
I will check how many designs failed in CI due to this PDN update. I will try to use some workaround to pass previous shuttle designs.

Its a request. Please keep me updated if there are any changes with respect to existing configuration variable.
Thanks @marwaneltoukhy  @kareefardi  @donn

PR #1182, opened 2022-07-03
Added missing -log to scrot_klayout call

Syndace:
This caused a crash when running with `TAKE_LAYOUT_SCROT`. All other usages of `scrot_klayout` seem to already set the `-log` arg correctly, just this occurrence was missed.


donn (approved):
Thanks for catching that.

PR #1261, opened 2022-08-11
Revamp Net Unbuffering

donn:
\+ `UNBUFFER_NETS` created, which takes a regular expression, which if matched, remove buffers from said net
~ `scripts/odbpy/remove_buffers.py` rewritten to be comprehensible
~ `remove_buffers_from_ports` -> `remove_buffers_from_nets`, now runs after every resizer step
~ `remove_*` python commands now only use regular expressions
\- Remove `DONT_BUFFER_PORTS` as the name does not reflect the proactive nature of buffer removal
\- Remove `remove_component` (singular)- unused
This is a temporary workaround until a proper "don't touch net" feature is available in OpenROAD.


donn:
i know he swapped those numbers i know it was 1216 one after magna carta etc

kareefardi (approved):

PR #1315, opened 2022-09-05
Need to read configuration before using QUIT_ON_MISMATCHES

antonblanchard:
When running Openlane with an updated PDK I see:

no such variable
    (read trace on "::env(QUIT_ON_MISMATCHES)")


donn:
This surfaces another bug when using a new version of OpenLane with an old version of a PDK (it won't find the variable STD_CELL_LIBRARY set when it needs it)

Let me tackle this in a different way.

PR #1316, opened 2022-09-05
Replace mismatches' environment variables with flags

donn:
~ `TEST_MISMATCHES` -> `./flow.tcl -test_mismatches <class>`
~ `QUIT_ON_MISMATCHES` o-> `./flow.tcl -ignore_mismatches`


antonblanchard:
@donn looks good, it makes much more sense to have them as options vs environment variables.

antonblanchard (approved):

Issue #1285, opened 2022-08-25
LEC fails with spm design example


mattvenn:
### Description

LEC fails with spm design

### Environment
```
python3 ./env.py issue-survey
Kernel: Linux v5.13.0-28-generic
Distribution: ubuntu 20.04
Python: v3.8.10 (OK)
Container Engine: docker v20.10.7 (OK)
OpenLane Git Version: f9b5781f5ef0bbdf39ab1c2bbd78be8db11b27f2
pip: INSTALLED
pip:venv: INSTALLED
---
PDK Version Verification Status: OK
---
Git Log (Last 3 Commits)

f9b5781 2022-07-01T16:04:31+02:00 Fix a bug with `-overwrite` (#1171) - Anton Blanchard -  (grafted, HEAD, tag: 2022.07.02_01.38.08)
```

### Reproduction Material

Add this line to config.tcl: set ::env(LEC_ENABLE) {1}
Then:
* make mount
* ./flow -design spm

No reproducible, openlane crashed too hard:

### Expected behavior

LEC should pass on all example designs.

### Logs
```
[INFO]: Running LEC: /openlane/designs/spm/runs/RUN_2022.08.25_15.20.58/results/placement/spm.resized.v.without_power_pins.v Vs. /openlane/designs/spm/runs/RUN_2022.08.25_15.20.58/results/synthesis/spm.v
[ERROR]: /openlane/designs/spm/runs/RUN_2022.08.25_15.20.58/results/placement/spm.resized.v.without_power_pins.v is not logically equivalent to /openlane/designs/spm/runs/RUN_2022.08.25_15.20.58/results/synthesis/spm.v.

    while executing
"logic_equiv_check -rhs $::env(PREV_NETLIST) -lhs $::env(CURRENT_NETLIST)"
    (procedure "run_resizer_design" line 19)
    invoked from within
"run_resizer_design"
    (procedure "run_placement" line 23)
    invoked from within
"run_placement"
    (procedure "run_placement_step" line 8)
    invoked from within
"[lindex $step_exe 0] [lindex $step_exe 1] "
    (procedure "run_non_interactive_mode" line 57)
    invoked from within
"run_non_interactive_mode {*}$argv"
    invoked from within
"if { [info exists flags_map(-interactive)] || [info exists flags_map(-it)] } {
    if { [info exists arg_values(-file)] } {
        run_file [file nor..."
    (file "./flow.tcl" line 389)

```

[...]

PR #1365, opened 2022-09-15
Fix LEC

emjunaid543:
This pull request solves the issue of LEC (logical equivalence check ) given in issue #1285.

### DIFF
--- a/scripts/yosys/logic_equiv_check.tcl
+++ b/scripts/yosys/logic_equiv_check.tcl
@@ -16,9 +16,7 @@ yosys -import
 if { [info exists ::env(FP_WELLTAP_CELL)] && $::env(FP_WELLTAP_CELL) ne ""} {
 	set well_tap_cell "$::env(FP_WELLTAP_CELL)"
 }
-set decap_cell_wildcard "$::env(DECAP_CELL)*"
-set fill_cell_wildcard "$::env(FILL_CELL)*"
-set buff_ef_cell_wildcare sky130_ef_sc_hd__decap*
+set all "$::env(CELL_PAD_EXCLUDE)"
 
 set vtop $::env(DESIGN_NAME)
 #set sdc_file $::env(SDC_FILE)
@@ -45,8 +43,11 @@ rmports
 if { [info exists well_tap_cell] } {
 	hierarchy -generate $well_tap_cell
 }
-hierarchy -generate $decap_cell_wildcard
-hierarchy -generate $fill_cell_wildcard
+
+foreach i $all {
+	hierarchy -generate $i
+	}
+
 splitnets -ports;;
 
 if { $::env(SYNTH_FLAT_TOP) } {
@@ -81,8 +82,11 @@ rmports
 if { [info exists well_tap_cell] } {
 	hierarchy -generate $well_tap_cell
 }
-hierarchy -generate $decap_cell_wildcard
-hierarchy -generate $fill_cell_wildcard
+
+foreach i $all {
+	hierarchy -generate $i
+	}
+
 splitnets -ports;;
 
 if { $::env(SYNTH_FLAT_TOP) } {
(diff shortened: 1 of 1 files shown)


scripts/yosys/logic_equiv_check.tcl
```diff
@@ -35,18 +36,19 @@ if { [info exists ::env(VERILOG_FILES_BLACKBOX)] } {
 	}
 }
 foreach lib $::env(LIB_SYNTH_COMPLETE_NO_PG) {
-	read_liberty -nooverwrite -lib -ignore_miss_dir -setattr blackbox $lib
```
donn, line 38: You're replacing the liberty file in the `foreach` loop with a different environment variable.

scripts/yosys/logic_equiv_check.tcl
```diff
@@ -70,18 +72,19 @@ if { [info exists ::env(VERILOG_FILES_BLACKBOX)] } {
...
```
donn, line 73: Ditto above

scripts/yosys/logic_equiv_check.tcl
```diff
@@ -107,7 +110,7 @@ if { [info exists ::env(VERILOG_FILES_BLACKBOX)] } {
...
```
donn, line 110: ditto above

scripts/yosys/logic_equiv_check.tcl
```diff
@@ -18,6 +18,7 @@ if { [info exists ::env(FP_WELLTAP_CELL)] && $::env(FP_WELLTAP_CELL) ne ""} {
 }
 set decap_cell_wildcard "$::env(DECAP_CELL)*"
 set fill_cell_wildcard "$::env(FILL_CELL)*"
+set buff_ef_cell_wildcare sky130_ef_sc_hd__decap*
```
donn, line 21: 1. Typo, you meant `wildcard`
2. This only works for sky130_fd_sc_hd. It's not PDK-generic.

donn (changes requested):

scripts/yosys/logic_equiv_check.tcl
emjunaid543: Because the liberty file provided by $lib environment is liberty file with no pg connections and you can only read it as black-box but there is no meaning of reading it as a black-box because when liberty file is read as black-box then during lec yosys can't figure out the logic of cells used in netlist and always fails the lec.

scripts/yosys/logic_equiv_check.tcl
emjunaid543: 1. yes it's a typo.
2. I'll try to make it generic.

scripts/yosys/logic_equiv_check.tcl
donn: Then you update `foreach lib $::env(LIB_SYNTH_COMPLETE_NO_PG) {` to `foreach lib $::env(LIB_SYNTH) {`. As it stands in your PR the for loop is meaningless and there is an inconsistency issue where you're looping over `LIB_SYNTH_COMPLETE_NO_PG` and then reading `LIB_SYNTH`.

scripts/yosys/logic_equiv_check.tcl
emjunaid543: set ::env(DECAP_CELL) {sky130_ef_sc_hd__decap_12 sky130_fd_sc_hd__decap_8 sky130_fd_sc_hd__decap_6 sky130_fd_sc_hd__decap_4 sky130_fd_sc_hd__decap_3}
The above env variable is taken from config.tcl generated during the run. As you can see it has all decap cells used in the design. Using the command given below:
**hierarchy -generate $decap_cell_wildcard** , where $decap_cell_wildcard = env(DECAP_CELL)
All decap cells should be removed from the design and this command is removing all decap cells present in **env(DECAP_CELL)** env variable but not removing "sky130_ef_sc_hd__decap_12". I have to declare it explicitly and then again use the above-provided command to eliminate this decap cell from design.
Do you know a way to get rid of this behavior?

scripts/yosys/logic_equiv_check.tcl
emjunaid543: Updated

donn:
Good call on CELL_PAD_EXCLUDE, but I just tested LEC_ENABLE=1 locally. It's still broken.

Issue #1328, opened 2022-09-06
`DIODE_INSERTION_STRATEGY` 4 `DIODE_PADDING`


kareefardi:
### Prompt
`DIODE_INSERTION_STRATEGY` `4` (and possibly `5`) isn't using padding specified by `DIODE_PADDING` during legalizing inserted diodes

### Proposal
All strategies should respect `DIODE_PADDING`

donn:
Moot as of #1686.

PR #1383, opened 2022-09-21
using `DIODE_PADDING` for `ins_diode_cells_4`

kareefardi:
(not the cleanest) workaround for https://github.com/The-OpenROAD-Project/OpenLane/issues/1328


donn:
Does this re-pad cells that have already been placed?

kareefardi:
It shouldn't as the cell are already placed. Unless that's the intention which would be weird. I will have to check anyway.

donn:
Superseded by #1686

PR #1395, opened 2022-09-26
Generate Timing Models

### LINKED ISSUE #171: Generate .lib files for hardened macros
As a start, this can be a dummy liberty file with basic information about the area and pins. A next step would be using OpenSTA to add more useful/accurate timing information into the file.

donn:
+ Add function to generate lib files based on OpenSTA's new `write_timing_model` command
+ Implement lib saving for both single and multi-corner STA
+ Add multi-corner SDF saving
~ Arrange RCX extraction and STA outputs under folders

---
Resolves #171

### DIFF
--- a/scripts/openroad/common/io.tcl
+++ b/scripts/openroad/common/io.tcl
@@ -160,7 +160,36 @@ proc write {args} {
     }
 
     if { [info exists ::env(SAVE_SDF)] } {
-        puts "Writing SDF to $::env(SAVE_SDF)..."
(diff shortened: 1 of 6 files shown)


tspyrou:
@donn it is essential that the end users know that they must perform model QA on this model as compared to the netlist before they trust it. If the netlist from which the model was generated is poorly buffered the model will not be accurate.

kareefardi (approved):

donn:
@tspyrou Agreed- what do you believe would be a good place for this information?

PR #1396, opened 2022-09-27
Fix `remove_buffers.py`

kareefardi:
+ add missing call for `getConstName()`
+ add the buffer output net name to the logging message
~ connect the load pin of cell driving the removed buffer to all cells that it is connected to. Previously, it was only connecting to the direct load of the removed buffer

### DIFF
--- a/scripts/odbpy/remove_buffers.py
+++ b/scripts/odbpy/remove_buffers.py
@@ -93,8 +93,7 @@ def get_io(cell: odb.dbInst):
     help="A regular expression matching all nets to remove.",
 )
 @click_odb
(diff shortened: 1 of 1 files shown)


scripts/odbpy/remove_buffers.py
```diff
@@ -30,22 +30,20 @@ def get_pin_name(pin: odb.dbITerm):
     return f"{cell_name}/{master_pin_name}"
 
 
-def get_sinks(net: odb.dbNet) -> List[odb.dbInst]:
```
donn, line 33: Why did you not update the type annotation?

donn (changes requested):

PR #1398, opened 2022-09-28
Support OpenROAD `set_dont_touch`

### LINKED ISSUE #1298: use set_dont_touch from openroad
### Prompt
Instead of manually removing buffer on specific nets utilize the recently added `set_dont_touch` from openROAD.

### Proposal
Instead of manually removing buffer on specific nets utilize the recently added `set_dont_touch` from openROAD.

donn:
\+ Add calls to `set_dont_touch` and then `unset_dont_touch` at the beginning and end of every resizer script respectively ~ Reorganize config variables
\+ Added undocumented variable `RSZ_USE_OLD_REMOVER` to continue to use the old `remove_buffers.py` script instead
\~ `LIB_RESIZER_OPT` -> `RSZ_LIB` (with translation behavior)
~ `UNBUFFER_NETS` -> `RSZ_DONT_TOUCH_RX` (with translation behavior)
~ Made timing models only get written after CTS by checking for `STA_PRE_CTS` as well

---

Resolves #1298


kareefardi (approved):

kareefardi:
We still need (possibly as a workaround) `remove_buffers.py` for some rare cases. The function of `remove_buffers.py` partially overlaps `set_dont_touch`.

maliberty:
Fwiw, OR has a remove_buffers command

kareefardi:
> Fwiw, OR has a remove_buffers command

It doesn't take any arguments and I think it removes all the buffers it can remove.

maliberty:
> > Fwiw, OR has a remove_buffers command
> 
> It doesn't take any arguments for which buffers to remove.

In the script I see:
```
    # This is a workaround for some situations where the resizer would buffer
    # analog ports.
```
if this is a bug it should be reported...

donn:
@kareefardi Are you going to approve this or?

kareefardi (approved):

kareefardi:
Following Mat's reply, I believe we should remove the comment that he is highlighting from the script.

PR #1418, opened 2022-10-10
Organize Magic Scripts

### LINKED ISSUE #1416: expose magic `-noblockage` during def read
### Prompt
def read `-noblockage` is not exposed for magic.

### Proposal
expose it. in the flow, we use blockages as routing obstructions for various reasons but it doesn't not necessarily mean that we want the mag (and dependent views like lef) to have these obstructions

### LINKED ISSUE #1417: expose magic `"-labels` during def read
### Prompt
magic flag isn't available.

### Proposal
expose the magic flag. from the docs:
> The "-labels" option to the "def read" command causes each net in the NETS and SPECIALNETS sections of the DEF file to be annotated with a label having the net name as the label text.

This is generally useful.

donn:
+ Add `proc erase_box` to OpenLane
+ Add two new config variables for magic def reads: `MAGIC_DEF_NO_BLOCKAGES` and `MAGIC_DEF_LABELS`
~ Magic scripts hierachically organized by input format
~ `erase_box.sh` deprecated

---
Resolves #1416
Resolves #1417


scripts/magic/erase_box.sh
```diff
@@ -1,29 +1,31 @@
...
 
-export PDK=sky130A
+Please update your interactive scripts to use the built-in OpenLane function
+"erase_box" instead.
+====================
+HEREDOC
+
+MAGIC_SCRIPTS_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
 
+export PDK=sky130A
```
kareefardi, line 22: is that needed here?

donn: That was the way it was in the original script, so yes.

kareefardi (approved):

scripts/report/report.py
```diff
@@ -679,7 +679,7 @@ def sta_report_extraction(
         filler_cells = tapcells + endcaps + diodes
 
         # LVS Total Errors
-        lvs_report = Artifact(rp, "logs", "signoff", f"{self.design_name}.lvs.lef.log")
+        lvs_report = Artifact(rp, "reports", "signoff", f"{self.design_name}.lvs.rpt")
```
mattvenn, line 682: nice

PR #1440, opened 2022-10-19
Renamed `cvc` to `cvc_rv` to avoid conflict

d-m-bailey:
Has not been tested. Not sure of all the intricacies of building a new docker image.

Renamed `docker/cvc/Dockerfile to `docker/cvc_rv/Dockerfile`.


donn:
I think #1440 could fix that presuming I could get it working

donn (approved):
Done: Though the PDK needs to be updated. Currently CVC_RV is not finding the efabless fill cells, so the check isn't running.

PR #1449, opened 2022-10-25
fix usage command in docs

yihuajack:
Correct Usage command in exploration scripts - Using OpenLane -> Regression & Exploration section


donn (approved):
Thanks for catching that!

PR #1453, opened 2022-10-26
Move flow control from atomic functions to flow functions

donn:
~ Move `RUN_FILL_INSERTION`, `RUN_TAP_DECAP_INSERTION`, `RUN_DRT` to the larger flow functions
~ `FILL_INSERTION` -> `RUN_FILL_INSERTION` w/ translation behavior
~ `TAP_DECAP_INSERTION` -> `RUN_TAP_DECAP_INSERTION` w/ translation behavior


kareefardi (changes requested):
1. `RUN_DRT` is added only in the routing function and has no default value.
2. Changes are not reflected in the docs.

donn:
1. [set ::env(RUN_DRT) 1](https://github.com/donn/openlane/blob/3b0505ca149d64523546ba44a974b94ce7b5fad7/configuration/general.tcl#L40)
2. The docs never specified whether the variables inhibit them in the flow or disable the function altogether.

kareefardi (approved):

PR #1461, opened 2022-10-31
`io_place` virtual pins + test design

ZeyadZaki:
- You can now use virtual pins in the pin_order configurations file by using `$<number_of_virtual_pins>`. 
- This functionality will give the user more flexibility while placing their i/o pins
- You can use virtual pins to place your actual pins closer together, further separate them or shift them towards a certain edge
- All changes were made only to the `scripts/odbpy/io_place.py` file
- There is an added design called `io_placer` which is a replica of the `caravel_upw` design except it uses the virtual pins concept in the `pin_order.cfg` file; you can run the flow on both of these designs and check the log file under `<run_path>/logs/floorplan/4-place_io.log` to see how the placement of the pins was changed.
- The algorithm uses the same logic that was used for placement before, so no change will be noticed if no virtual pins were used. 
- An example of how to use the feature; in the pin_order.cfg file if you wrote:
```
#N
x
$20
y
```
The algorithm will consider that there are 22 pins on the north edge, will place x in the first assigned pin slot and y in the last slot and there will be 20 empty slots in between them.


maliberty:
Why is OL maintaining a separate IO pin placer from the one in OR?  If there any capabilities needed in OR then please file an issue (or PR).

donn:
@maliberty Would anyone from the OpenROAD team be willing to put in the time and effort to port all existing pin placement files to be OpenROAD-compatible? Do note this includes Caravel, DFFRAM and a litany of dependent projects.

scripts/odbpy/io_place.py
```diff
@@ -35,52 +35,75 @@ def grid_to_tracks(origin, count, step):
     return tracks
 
 
-def equally_spaced_sequence(pin_count, possible_locations):
+def equally_spaced_sequence(side, side_pin_placement, possible_locations):
```
donn, line 38: Kindly read the contributors' guide. All Python code has to be formatted and linted before being merged into OpenLane. This file does not pass these checks after your modifications.

scripts/odbpy/io_place.py
```diff
@@ -287,25 +310,34 @@ def io_place(
...
-                            ". Those are",
-                            bterm_regex_map[bterm],
-                            "and",
-                            regex,
-                        )
-                        sys.exit(os.EX_DATAERR)
-                    bterm_regex_map[bterm] = regex
-                    pin_placement[side].append(bterm)  # to maintain the order
+            if regex[0] == "$": # Sign of Virtual Pins
+                if regex.split("$")[1].isdigit(): # Check that only numbers are followed by the dollar sign
```
donn, line 314: It is best to try to execute int(regex[1:]) than catch an exception rather than check the data in advance. See https://devblogs.microsoft.com/python/idiomatic-python-eafp-versus-lbyl/

scripts/odbpy/io_place.py
```diff
@@ -287,25 +310,34 @@ def io_place(
...
+            else:
+                regex += "$"  # anchor
+                for bterm in bterms:
+                    # if a pin name matches multiple regexes, their order will be
+                    # arbitrary. More refinement requires more strict regexes (or just
+                    # the exact pin name).
+                    pin_name = bterm.getName()
+                    if re.match(regex, pin_name) is not None:
+                        if bterm in bterm_regex_map:
+                            print(
```
donn, line 330: Please use a formatted string.

scripts/odbpy/io_place.py
```diff
@@ -35,52 +35,75 @@ def grid_to_tracks(origin, count, step):
...
+            current_track += tracks_per_pin
+    else: # There are virtual pins
+        for i in range(len(side_pin_placement)):
+            if not isinstance(side_pin_placement[i], int): # We have an actual pin
+               result.append(possible_locations[current_track]) 
+               current_track += tracks_per_pin
+            else: # Virtual Pins, so just leave their needed spaces 
+                current_track += tracks_per_pin * side_pin_placement[i]
+        side_pin_placement = [pin for pin in side_pin_placement if not isinstance(pin, int)] # Remove the virtual pins from the side_pin_placement list
+    VISUALIZE_PLACEMENT = True
```
donn, line 82: Is there any particular reason why you changed the VISUALIZE_PLACEMENT block into something completely different instead of simply printing statistics in a separate block?

scripts/odbpy/io_place.py
donn, line 38: Kindly read the contributors' guide. All Python code has to be formatted and linted before being merged into OpenLane. This file does not pass these checks after your modifications.

donn (changes requested):

designs/io_placer/config.tcl
```diff
@@ -0,0 +1,128 @@
+# SPDX-FileCopyrightText: 2020 Efabless Corporation
```
donn, line 1: Also... why config.tcl? We'd moved on to .json for almost all designs. See `manual_macro_placement_test` for an example.

scripts/odbpy/io_place.py
ZeyadZaki: Done.

scripts/odbpy/io_place.py
ZeyadZaki: Now the statistics are printed outside the block as you requested, I just kept the VISUALIZE_PLACEMENT = True to print out the used_track_indices which is important for understanding how the virtual pins actually works.

scripts/odbpy/io_place.py
ZeyadZaki: Done.

designs/io_placer/config.tcl
ZeyadZaki: I used the config.tcl as it was an exact replica from the caravel_upw design which still used tcl, however I changed the config file to JSON in this commit.

scripts/odbpy/io_place.py
ZeyadZaki: I'm not sure what you need here this peace of code was already there in the script, nothing changed by me here

ZeyadZaki: Done.

donn:
Lint...

donn (approved):

PR #1485, opened 2022-11-11
`LIB_SYNTH_COMPLETE` -> `LIB_SYNTH` in some OpenROAD scripts

yupferris:
When using [asap7](https://github.com/The-OpenROAD-Project/asap7) (using [this version of volare](https://github.com/efabless/volare/pull/29) to install/activate), `LIB_SYNTH_COMPLETE` specifies several files, instead of a single one, like many commands in the flow scripts expect. To work around this, we can specify `LIB_SYNTH` instead, which, after `prep` in all.tcl is invoked, refers to the output of Liberty file merging, which should contain everything we need to continue in the flow.

This gets designs through the flow until a crash in floorplanning detailed [here](https://github.com/The-OpenROAD-Project/OpenLane/issues/1486).

### DIFF
--- a/scripts/openroad/common/io.tcl
+++ b/scripts/openroad/common/io.tcl
@@ -38,7 +38,7 @@ proc read_libs {args} {
         keys {-override}\
         flags {-multi_corner}
 
-    set libs $::env(LIB_SYNTH_COMPLETE)
+    set libs $::env(LIB_SYNTH)
 
     if { [info exists keys(-override)] } {
         set libs $keys(-override)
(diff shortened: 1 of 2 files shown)


scripts/tcl_commands/all.tcl
```diff
@@ -736,7 +736,7 @@ proc prep {args} {
         merge_lib\
             -output $::env(LIB_SYNTH_MERGED)\
             -name $::env(PDK)_merged\
-            -inputs {*}$::env(LIB_SYNTH_COMPLETE)
```
donn, line 739: This is actually the only line I'm concerned about- the rest looks fine

yupferris:
After @donn 's last comment and checking this again, I'm more confident that these changes are correct, so promoting this from draft status.

donn (approved):

PR #1498, opened 2022-11-16
Workaround hacks 2

yupferris:


donn:
As this is primarily for ASAP7- we've decided to just keep that for OpenLane 2, where this issue has been fixed. Thank you for your contribution.

Issue #1497, opened 2022-11-16
Running IO Placement stage fails


vijayank88:
### Description

This is MPW CI [design](https://github.com/The-OpenROAD-Project/caravel_user_project/blob/SultanShadow/APPROX_MULT/openlane/user_proj_example/pin_order.cfg), follows pin order name with xyz.*

As its MPW6 design, designer specified only few pins in specific side and left some of pins to place randomly.

With latest OpenLane setup its made compulsory to define manual placement of IO pins or go with auto pin placement technique. To skip IO placement error for un-defined pin placement used `set ::env(FP_IO_UNMATCHED_ERROR) 0`.

After the following PR https://github.com/The-OpenROAD-Project/OpenLane/pull/1461 facing error again. Help me how to resolve the error.

### Expected Behavior

Flow pass IO placement stage.

### Environment report

```shell
Kernel: Linux v3.10.0-1160.49.1.el7.x86_64
Distribution: centos 7
Python: v3.6.8 (OK)
Container Engine: docker v20.10.16 (OK)
OpenLane Git Version: 1da2eec3d9b5a3612cb026cf9060cb7954e4c24a
pip: INSTALLED
python-venv: INSTALLED
---
PDK Version Verification Status: OK
---
Git Log (Last 3 Commits)

1da2eec 2022-11-16T01:37:27+02:00 [BOT] Update netgen (#1452) - Openlane Bot -  (HEAD -> master, tag: 2022.11.16, origin/master, origin/HEAD)
77fd531 2022-11-15T13:32:13+02:00 Revert #1485 (#1492) - Mohamed Gaber -  ()
dbb2487 2022-11-15T13:05:26+02:00 [BOT] Update PDK (#1455) - Openlane Bot -  ()
---
Git Remotes

origin	https://github.com/The-OpenROAD-Project/OpenLane.git (fetch)
origin	https://github.com/The-OpenROAD-Project/OpenLane.git (push)
```

### Reproduction material

Unzip `APPROX_MULT.zip` 
[APPROX_MULT.zip](https://github.com/The-OpenROAD-Project/OpenLane/files/10020308/APPROX_MULT.zip)
 and place it `OpenLane/designs/`

%cd OpenLane
%make mount
%./flow.tcl -design APPROX_MULT -verbose 99

### Relevant log output

```shell
[INFO]: Running IO Placement (log: logs/floorplan/4-place_io.log)...
[...]

PR #1503, opened 2022-11-18
fix return values in `equally_spaced_sequence`

kareefardi:
trying to address https://github.com/The-OpenROAD-Project/OpenLane/issues/1497


donn (approved):
LGTM

PR #1509, opened 2022-11-20
Document Error in JSON Config Files

donn:
+ Add error message and document corner case where `ref::` is used with a non-string


kareefardi (approved):

designs/spm/config.json
```diff
@@ -2,6 +2,7 @@
     "DESIGN_NAME": "spm",
     "VERILOG_FILES": "dir::src/*.v",
     "CLOCK_PERIOD": 10,
+    "TEST_FLOAT_REF": "expr::$CLOCK_PERIOD",
```
proppy, line 5: do you really want to commit this? (or should it be a test case)

PR #1511, opened 2022-11-24
Fix LEC

antonblanchard:
Not sure how you want to fix this, but this patch partially undoes a recent change:

```
-    if { [catch {exec $::env(SYNTH_BIN) -c $::env(SCRIPTS_DIR)/yosys/logic_equiv_check.tcl -l $log |& tee $::env(TERMINAL_OUTPUT)} ]}
 {
+    if { [run_yosys_script $::env(SCRIPTS_DIR)/yosys/logic_equiv_check.tcl -indexed_log $log]} {
```

There is no output from `run_yosys_script`, so it doesn't make sense to use it in a conditional. The current failure:

```
expected boolean value but got ""
    while executing
"if { [run_yosys_script $::env(SCRIPTS_DIR)/yosys/logic_equiv_check.tcl -indexed_log $log]} {
        puts_err "$::env(LEC_LHS_NETLIST) is not logicall..."
    (procedure "logic_equiv_check" line 31)
    invoked from within
```

### DIFF
--- a/scripts/tcl_commands/synthesis.tcl
+++ b/scripts/tcl_commands/synthesis.tcl
@@ -210,7 +210,7 @@ proc logic_equiv_check {args} {
     set rhs_rel [relpath . $::env(LEC_RHS_NETLIST)]
     puts_info "Running LEC: '$lhs_rel' vs. '$rhs_rel' (log: [relpath . $log])..."
 
-    if { [run_yosys_script $::env(SCRIPTS_DIR)/yosys/logic_equiv_check.tcl -indexed_log $log]} {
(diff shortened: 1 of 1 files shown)


kareefardi:
@antonblanchard I think the intent was 
```
    if { [catch {run_yosys_script $::env(SCRIPTS_DIR)/yosys/logic_equiv_check.tcl -indexed_log $log}] } {
```
Does that work for you ?

donn (approved):
I changed it to reflect Karim's suggestion- `run_yosys_script` allows reproducibles to be created.

PR #1520, opened 2022-11-28
Fix processing of pdk_dir:: and scl_dir:: in JSON

antonblanchard:


donn (approved):
Oh. Thank you for catching that.

PR #1521, opened 2022-11-28
Fix manual macro placement on GF180

antonblanchard:
We assume database units per micron is 1000, but on GF180 it is 2000.


antonblanchard:
before:
def_database_microns_bug

after (ignore the overlapping macros on the left, it was a mistake in my calculations):
def_fixed

donn (approved):

PR #1524, opened 2022-11-28
Fix forgotten openroad binary env variable

ekliptik:
OpenLane cannot be used without this PR outside the Docker when OpenROAD is not on PATH


ekliptik:
Somehow, my working directory couldn't pull from master. This has apparently already been fixed

PR #1528, opened 2022-11-29
Makefile: disable seccomp to enable openroad ui

### LINKED ISSUE #1149: Is there a way to visualize the DEF file into the OpenROAD gui?
Hi, I come from the ORFS flow and I found pretty useful the gui. I just finished synthetizing a design and i would like to visualize the gds file with the gui. The documentation doesn't seems to talk about OpenROAD commands etc.

proppy:
Resolves #1149

As described in https://docs.docker.com/engine/security/seccomp/#run-without-the-default-seccomp-profile

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,7 +31,7 @@ endif
 # Allow using GUIs
 UNAME_S = $(shell uname -s)
 ifeq ($(UNAME_S),Linux)
-DOCKER_OPTIONS += -e DISPLAY=$(DISPLAY) -v /tmp/.X11-unix:/tmp/.X11-unix -v $(HOME)/.Xauthority:/.Xauthority --network host
+DOCKER_OPTIONS += -e DISPLAY=$(DISPLAY) -v /tmp/.X11-unix:/tmp/.X11-unix -v $(HOME)/.Xauthority:/.Xauthority --network host --security-opt seccomp=unconfined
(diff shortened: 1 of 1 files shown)


donn:
Are we sure this allows OpenLane to run in environments where users are using rootless Docker and/or are otherwise unprivileged?

proppy:
I'm not sure if the current set of flags works in rootless docker (https://docs.docker.com/engine/security/rootless/#known-limitations point out that host networked is namespaced for example).

maliberty:
@donn @proppy  what needs to happen to resolve this?

proppy:
I don't think OpenLane currently run in rootless, so the change would make their life worst in my opinion.

Maybe we could file a separate bug about testing/ fixing rootless for OpenLane if we merge this?

donn:
Rootless works, still need to test podman though...

donn (approved):

maliberty:
@mattvenn you can't pass the odb file as a command line argument.  You need to pass a tcl script to OR that loads your design.  That's the problem in your first attempt.

For the 2nd try I get the same.  It seems something has broken and I'll see if I can figure it out.

maliberty:
@mattvenn It appears that with -gui the design must already be compiled and so you must be using a valid tag to reference that run.  I find this works:
```
./flow.tcl -design spm -tag foo
./flow.tcl -design spm -tag foo -gui
```

PR #1566, opened 2022-12-11
expose magic polygon subcells

kareefardi:
\+ add configuration variable `MAGIC_GDS_POLYGON_SUBCELLS`
\~ minor formatting changes to other variables documentation

During caravel gf180 development. This greatly enhanced the performance of generating the chip's gds with magic


RTimothyEdwards:
I should probably review this.  The "gds polygon subcells" will have a large impact on _reading_ the foundry GDS, specifically for the I/O corner cell layout which has large structures at 45 degree angles with narrow separation.  It had better have zero impact on _writing_ GDS, or you're doing something wrong.

RTimothyEdwards:
Offhand, I'd say that the issue is that `EXTRA_GDS_FILES` should _only_ contain cells that are not in the PDK library and for which a .mag view does not exist.

kareefardi:
@RTimothyEdwards It is the case for caravel that extra gds doesn't contain anything from the pdk.  
```
set ::env(EXTRA_GDS_FILES) "\
    $gds_root/copyright_block.gds \
    $gds_root/open_source.gds \
    $gds_root/user_id_textblock.gds \
    $gds_root/caravel_logo.gds \
    $gds_root/caravel_motto.gds \
    $gds_root/caravel_power_routing.gds \
    $gds_root/chip_io.gds \
    $gds_root/caravel_core.gds"
```
About specifically using `gds polygon subcells`, I was told that the following statements impacted the performance of gds generation:
```
gds ordering on
gds polygon subcells true
gds flatten true
gds flatglob *_CDNS_*`
```
and only `gds polygon subcells true` had an impact

kareefardi:
Another enhancement we can do is have the ability to load mag files. Would that be simply achieved by iteration of a list mag file and called `load <file>` ?

donn:
@kareefardi Do I want to merge this now or wait?

RTimothyEdwards:
@kareefardi :  The problem is that this flow is not designed for reading in `chip_io` from GDS.  However, as you are running the above commands with the `gds readonly true` option set, then it should be okay to read in with `polygon subcells true`.  However, it would be preferable to let magic generate output for all layout except for the core, as that is the way the flow was designed to work.  But in the above script, the inefficiency caused by pulling all the GDS separately should not be much of an issue.

RTimothyEdwards:
@kareefardi :  Meaning that I'm okay with the script above, as long as we can confirm that the final GDS is correct.

kareefardi:
@RTimothyEdwards I see. I will generate layouts using both methods and compare the layouts.

kareefardi:
@donn We can potentially wait till I verify the layout generated. Or we can have it merged now and remove it later if it proves to be dangerous.

kareefardi:
@donn The comparison was found to be ok. I think we should merge this

kareefardi:
To be more verbose, there were differences however they originated from the source gds files not because of the magic command

donn (approved):

PR #1576, opened 2022-12-14
Improve `PL_BASIC_PLACEMENT` Documentation

xobs:
This ought to help clear up some confusion about what the variable actually does, why someone might want to turn it on, and why it probably should be off.


xobs:
I'm still having trouble understanding exactly why it exists, and looking on Slack it seems like this feature tends to trip people up more than it helps them.

maliberty:
I haven't looked at it directly but I've been told that for very small designs OR sometimes has trouble and basic placement will get you a result (not necessarily optimal but sufficient for trivial designs).  I encourage anyone who has such an issue to report it as an OR bug to be fixed instead.

kareefardi:
@maliberty can you point me to that discussion or designs if possible?

maliberty:
@kareefardi @mattvenn had a recent issue so you might check w/him.

donn (approved):

Issue #1574, opened 2022-12-14
[ERROR DPL-0036] Detailed placement failed -- error is not actionable


xobs:
### Description

When synthesizing a design, OpenLane produces the error `[ERROR DPL-0036] Detailed placement failed.`. It does not say why placement failed, nor does it explain how to address the error. It also doesn't report how to iterate to address the error, which is a problem when synthesis takes a lot of time.

### Expected Behavior

When a step fails, it should report *why* the step failed. Ideally it should also give suggestions on how to fix the issue.

### Environment report

```shell
user@Ondo:/opt/Si/work/caravel-gf180-pio/deps/openlane_src$ python3 ./env.py issue-survey
Kernel: Linux v5.15.74.2-microsoft-standard-WSL2
Distribution: ubuntu 22.04
Python: v3.10.6 (OK)
Container Engine: docker v20.10.21 (OK)
OpenLane Git Version: cb59d1f84deb5cedbb5b0a3e3f3b4129a967c988
pip: INSTALLED
python-venv: INSTALLED
---
PDK Version Verification Status: OK
---
Git Log (Last 3 Commits)

cb59d1f 2022-11-18T18:42:38+02:00 fix return values in `equally_spaced_sequence` (#1503) - Kareem Farid -  (grafted, HEAD, tag: 2022.11.19)
---
Git Remotes

origin  git@github.com:The-OpenROAD-Project/OpenLane (fetch)
origin  git@github.com:The-OpenROAD-Project/OpenLane (push)
user@Ondo:/opt/Si/work/caravel-gf180-pio/deps/openlane_src$
```

### Reproduction material

[issue_reproducible.tar.gz](https://github.com/The-OpenROAD-Project/OpenLane/files/10224306/issue_reproducible.tar.gz)

The repository is available at https://github.com/xobs/caravel-gf180-pio

### Relevant log output

```shell
[INFO]: Running Placement Resizer Design Optimizations (log: ../opt/Si/work/caravel-gf180-pio/openlane/wb_pio/runs/22_12_14_12_14/logs/placement/8-resizer.log)...
  [ERROR]: during executing openroad script /openlane/scripts/openroad/resizer.tcl
[ERROR]: Log: ../opt/Si/work/caravel-gf180-pio/openlane/wb_pio/runs/22_12_14_12_14/logs/placement/8-resizer.log
[ERROR]: Last 10 lines:
[INFO DPL-0035]  _075635_
[...]

PR #1577, opened 2022-12-14
scripts: yosys: generate readable names for instances

xobs:
When synthesizing Verilog using Yosys, internal, 6-digit codes are used for the resulting net names. This means that most nets have unhelpful names such as `_000019_`.

Use the `autoname` step prior to writing verilog in order to generate usable names.

### DIFF
--- a/scripts/yosys/synth.tcl
+++ b/scripts/yosys/synth.tcl
@@ -335,6 +335,17 @@ proc run_strategy {output script strategy_name {postfix_with_strategy 0}} {
     tee -o "$::env(synth_report_prefix).$strategy_escaped.chk.rpt" check
     tee -o "$::env(synth_report_prefix).$strategy_escaped.stat.rpt" stat -top $::env(DESIGN_NAME) -liberty [lindex $::env(LIB_SYNTH_NO_PG) 0]
 
+    if { [info exists ::env(SYNTH_AUTONAME)] && $::env(SYNTH_AUTONAME) } {
+        # Generate public names for the various nets, resulting in very long names that include
+        # the full heirarchy, which is preferable to the internal names that are simply
+        # sequential numbers such as `_000019_`. Renamed net names can be very long, such as:
+        #     manual_reset_gf180mcu_fd_sc_mcu7t5v0__dffq_1_Q_D_gf180mcu_ \
+        #     fd_sc_mcu7t5v0__nor3_1_ZN_A1_gf180mcu_fd_sc_mcu7t5v0__aoi21_ \
+        #     1_A2_A1_gf180mcu_fd_sc_mcu7t5v0__nand3_1_ZN_A3_gf180mcu_fd_ \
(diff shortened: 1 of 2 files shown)


xobs:
This helps to address some of the issues in https://github.com/The-OpenROAD-Project/OpenLane/issues/1574

kareefardi:
@xobs This is failing CI. I am checking it. 
I like the idea though, how about we wrap the command around a variable similar to `SYNTH_BUFFERING`, for instance ?

xobs:
What about calling it `SYNTH_AUTONAME`? Should it default to `true` or `false`?

What would be the case where you'd want autogenerated names? The biggest reason I could see is if there's a tool that doesn't support names with characters such as `[`...

maliberty:
I tried it and the names it generates are ok but some are overly long.

marwaneltoukhy:
While debugging the LVS issue in the APU (looks like this fix messed with something in the def/lef), I found it extremely hard to debug due to the long net names, so I'm concerned with the impact of this fix, and not sure if it is useful.

xobs:
I believe I've managed to add a new variable `SYNTH_AUTONAME` that enables this feature, which defaults to not running.

scripts/yosys/synth.tcl
```diff
@@ -335,6 +335,13 @@ proc run_strategy {output script strategy_name {postfix_with_strategy 0}} {
     tee -o "$::env(synth_report_prefix).$strategy_escaped.chk.rpt" check
     tee -o "$::env(synth_report_prefix).$strategy_escaped.stat.rpt" stat -top $::env(DESIGN_NAME) -liberty [lindex $::env(LIB_SYNTH_NO_PG) 0]
 
+    if { [info exists ::env(SYNTH_AUTONAME)] && $::env(SYNTH_AUTONAME) } {
+        # Generate public names for the various nets, resulting in very long names that include
+        # the full heirarchy, which is preferable to the internal names that are simply
+        # sequential numbers such as `_000019_`.
```
proppy, line 341: should we give an example of what the name looks like here too?

proppy (approved):

scripts/yosys/synth.tcl
xobs: Sure, I've added that to the comment. It does get pretty gnarly, and appears to crash magic.

Note that klayout also doesn't work, but that's because OpenLane is generating invalid arguments to klayout -- klayout works fine if you pass it the correct arguments.

kareefardi:
@RTimothyEdwards Is there a restriction on nets/instances name length in netgen or a known issue related to that ?

xobs:
I added an example name to the comment.

Note that OpenLane is currently unable to use this because of two unrelated issues:

* https://github.com/The-OpenROAD-Project/OpenLane/issues/1593
* https://github.com/The-OpenROAD-Project/OpenLane/issues/1587

Dolu1990:
Ahhhhh, i was realy about to open a new issue about that same case on yosys XD
SYNTH_AUTONAME => Thanks :D

PR #1601, opened 2023-01-02
padringer fixes

kareefardi:
`padringer.py`:
\+ Add output `ODB` and optional `DEF`
\+ Add `--odb-lef` which is the lef file to included with `ODB` output file.
\+ `--width` and `--height` are no longer required as they can be fetched from the config file
\+ `--input-lef` and `--padframe-config` click options are now multiple 
others:
\+ Add `FP_PADFRAME_CFG` which points to configuration file passed to `padringer.py`
\+ Add `GPIO_PADS_PREFIX` to parameterize prefix passed to `padringer.py`
\+ Add `padframe_gen_batch` to capture all of the above in a simple wrapper


scripts/openroad/read_write.tcl
```diff
@@ -0,0 +1,16 @@
+# Copyright 2020-2022 Efabless Corporation
```
donn, line 1: This is an exact duplicate of scripts/openroad/write_views.tcl

donn (changes requested):

scripts/openroad/read_write.tcl
kareefardi: removed

donn (approved):

PR #1608, opened 2023-01-07
KLayout GDSII updates

kareefardi:
\+ Add `KLAYOUT_DEF_LAYER_MAP` for DEF/LEF mapping in klayout. `.lyt` is not sufficient to map a pin shape to pin and metal for example. 
\+ Add `KLAYOUT_XOR_IGNORE_LAYERS` see https://github.com/RTimothyEdwards/open_pdks/pull/347 (default: empty) 
\~ Enable `QUIT_ON_XOR_ERROR` by default


donn (approved):
LGTM- great work

PR #1612, opened 2023-01-10
Update KLayout + Script Tweaks

donn:
~ Klayout -> 0.28.2 to fix an issue with parsing LEF files
~ Tweaks to KLayout scripts


donn:
@kareefardi Before you ask: `read_lef_with_def` just means it will automatically look for a LEF file in the same folder, and if it is found, it is also loaded. This causes issues so I prefer to just list the LEF file(s) explicitly.

kareefardi (approved):

Issue #1600, opened 2023-01-01
Verilog `generate` does not preserve power connections


mole99:
### Description

When instantiating multiple macros via the Verilog `generate` statement the power connections are lost somewhere in the flow and thus the macros are not connected to power.

I had a look at the resulting file `tmp/signoff/25-user_project_wrapper.pnl.v`. The macro I had instantiated normally is connected to `vccd1`/`vssd1`, the other four instances via the `generate` statement are not connected to `vccd1`/`vssd1`.

Example code snippet:

```
generate
    genvar i;

    for (i = 0; i < NUM_INSTANCES; i++) begin : memory

        sky130_sram_2kbyte_1rw1r_32x512_8 sram (
            `ifdef USE_POWER_PINS
            .vccd1(vccd1),  // User area 1 1.8V power
            .vssd1(vssd1),  // User area 1 digital ground
            `endif
            // Port 0: RW
            .clk0  (wb_clk_i),
            .csb0  (la_data_in[0]),
            .web0  (la_data_in[1]),
            .wmask0(la_data_in[5:2]),
            .addr0 (la_data_in[14:6]),
            .din0  (la_data_in[46:15]),
            .dout0 (select_dout0[i*DATA_WIDTH+:DATA_WIDTH]),
            // Port 1: R
            .clk1 (wb_clk_i),
            .csb1 (la_data_in[47]),
            .addr1(la_data_in[79:48]),
            .dout1(select_dout1[i*DATA_WIDTH+:DATA_WIDTH])
        );
        
    end
endgenerate
```

### Expected Behavior

The macros should be connected to power.

### Environment report

```shell
$ python3 ./env.py issue-survey
Kernel: Linux v6.0.0-6-amd64
Distribution: debian 
Python: v3.10.9 (OK)
Container Engine: docker v20.10.21+dfsg1 (OK)
OpenLane Git Version: cb59d1f84deb5cedbb5b0a3e3f3b4129a967c988
pip: INSTALLED
python-venv: INSTALLED
---
PDK Version Verification Status: MISMATCH
The version of open_pdks used in building the PDK does not match the version OpenLane was tested on (installed: 0059588eebfc704681dc2368bd1d33d96281d10f, tested: 3af133706e554a740cfe60f21e773d9eaa41838c)
This may introduce some issues. You may want to re-install the PDK by invoking `make pdk`.
---
Git Log (Last 3 Commits)

[...]

PR #1622, opened 2023-01-15
`write_verilog` from a `DEF` file

kareefardi:
~ allow write_verilog to have an input def file 
~ allow reading in def/lef for openroad scripts

More on the PR:

Addresses a bug in `write_powered_verilog` where `write_verilog` didn't read the def produced by `write_powered_def`. `write_powered_def` proved to be useful as it parses a yosys netlist where(supposedly) all power connections are defined. If due to a misconfiguration or other reasons, openroad creates faulty power connections, the openroad gl netlist will also have the same faulty power connections resulting in misleading lvs results. The mentioned behavior was surfaced in https://github.com/The-OpenROAD-Project/OpenLane/issues/1600

### DIFF
--- a/scripts/openroad/common/io.tcl
+++ b/scripts/openroad/common/io.tcl
@@ -86,7 +86,7 @@ proc read {args} {
         keys {-override_libs}\
         flags {-multi_corner_libs}
 
-    if { [info exists ::env(IO_READ_DEF)] } {
+    if { [info exists ::env(IO_READ_DEF)] && $::env(IO_READ_DEF) } {
         if { [ catch {read_lef $::env(MERGED_LEF)} errmsg ]} {
             puts stderr $errmsg
             exit 1
--- a/scripts/openroad/common/io.tcl
+++ b/scripts/openroad/common/io.tcl
@@ -86,7 +86,17 @@ proc read {args} {
         keys {-override_libs}\
         flags {-multi_corner_libs}
 
-    if {[catch {read_db $::env(CURRENT_ODB)} errmsg]} {
+    if {
+        [catch {
+            if { [info exists ::env(DEF_IN)] } {
+                puts "Reading def $::env(DEF_IN)"
+                read_lef $::env(MERGED_LEF)
+                read_def $::env(DEF_IN)
+            } else {
+                puts "Reading odb $::env(CURRENT_ODB)"
+                read_db $::env(CURRENT_ODB)
+            }
+        } errmsg]} {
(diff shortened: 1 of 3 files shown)


scripts/openroad/common/io.tcl
```diff
@@ -86,7 +86,17 @@ proc read {args} {
         keys {-override_libs}\
         flags {-multi_corner_libs}
 
-    if {[catch {read_db $::env(CURRENT_ODB)} errmsg]} {
+    if {
+        [catch {
+            if { [info exists ::env(DEF_IN)] } {
+                puts "Reading def $::env(DEF_IN)"
```
donn, line 92: Remove debug vestiges?

scripts/utils/utils.tcl
```diff
@@ -486,6 +486,10 @@ proc run_tcl_script {args} {
     set index 1
     set name $::env(DESIGN_NAME)
 
+    if { [info exists flag_map(-def_in)] } {
```
donn, line 489: Why not make IO_READ_DEF a flag and then just switch to CURRENT_DEF? Wouldn't that be cleaner?

scripts/openroad/common/io.tcl
kareefardi: I left "Reading odb" because openroad doesn't print the input odb file while it does print the def and lef

scripts/utils/utils.tcl
kareefardi: I made some changes. Are they as you would expect ?

scripts/openroad/common/io.tcl
```diff
@@ -86,7 +86,16 @@ proc read {args} {
...
         flags {-multi_corner_libs}
 
-    if {[catch {read_db $::env(CURRENT_ODB)} errmsg]} {
+    if {
+        [catch {
+            if { [info exists ::env(IO_READ_DEF)] } {
+                read_lef $::env(MERGED_LEF)
+                read_def $::env(CURRENT_DEF)
+            } else {
+                puts "Reading odb $::env(CURRENT_ODB)"
```
donn, line 95: debug vestige?

kareefardi: I left "Reading odb" because openroad doesn't print the input odb file while it does print the def and lef

donn (approved):
Tweaked a bit, but LGTM. Will merge.

kareefardi:
@donn do you need anything else here?

PR #1626, opened 2023-01-16
Fix loading design configs again

kareefardi:
\- revert previous "fix"

\~ only substitute glob when there is a match.
\~ wrap section that parses each variable in try and catch so that the exception message contains both key and value. Previously, some exceptions thrown due to value errors were captured at the main function scope and hence losing the ability to find at which key the exception occurred. This is not a mandatory change and can be removed if not desirable.

### DIFF
--- a/scripts/config/tcl.py
+++ b/scripts/config/tcl.py
@@ -254,12 +254,11 @@ def process_string(value: str, state: State) -> str:
             full_abspath = os.path.abspath(value)
 
             # Resolve globs for paths that are inside the exposed directory
-            if value.startswith("/"):
-                if full_abspath.startswith(found):
-                    files = glob.glob(full_abspath)
-                    if files:
-                        files_escaped = [file.replace("$", r"\$") for file in files]
-                        value = " ".join(files_escaped)
+            if value.startswith("/") and full_abspath.startswith(found):
+                files = glob.glob(full_abspath)
+                if files:
+                    files_escaped = [file.replace("$", r"\$") for file in files]
+                    value = " ".join(files_escaped)
         except KeyError:
             raise InvalidConfig(
                 f"Referenced variable '{reference_variable}' not found."
--- a/scripts/config/tcl.py
+++ b/scripts/config/tcl.py
@@ -294,7 +294,7 @@ def process_config_dict_recursive(config_in: Dict[str, Any], state: State):
         withhold = False
         try:
             if not isinstance(key, str):
-                raise InvalidConfig(f"must be a string.")
+                raise InvalidConfig("must be a string")
             if isinstance(value, dict):
                 withhold = True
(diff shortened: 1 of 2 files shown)


donn:
I must admit I don't entirely see the value in reporting a failed glob as an exception. Why not just print a warning?

scripts/config/tcl.py
```diff
@@ -254,10 +254,15 @@ def process_string(value: str, state: State) -> str:
...
-                value = " ".join(files_escaped)
+            if value.startswith("/"):
+                if full_abspath.startswith(found):
+                    files = glob.glob(full_abspath)
+                    if not files:
+                        raise InvalidConfig(f"No such file {full_abspath}")
+                    files_escaped = [file.replace("$", r"\$") for file in files]
+                    value = " ".join(files_escaped)
+                else:
+                    raise InvalidConfig(f"Invalid path: {full_abspath}")
```
donn, line 265: That's not an invalid path... That's literally any path that would ascend from the PDK or SCL.

scripts/config/tcl.py
```diff
@@ -254,10 +254,15 @@ def process_string(value: str, state: State) -> str:
...
             # Resolve globs for paths that are inside the exposed directory
-            if value.startswith("/") and full_abspath.startswith(found):
-                files = glob.glob(full_abspath)
-                files_escaped = [file.replace("$", r"\$") for file in files]
-                value = " ".join(files_escaped)
+            if value.startswith("/"):
+                if full_abspath.startswith(found):
+                    files = glob.glob(full_abspath)
+                    if not files:
+                        raise InvalidConfig(f"No such file {full_abspath}")
```
donn, line 261: Warning?

donn (approved):

scripts/config/tcl.py
donn, line 265: Not an invalid path, that's every path that would ascend...

scripts/config/tcl.py
donn, line 261: Why not just a warning? The glob thing is best-effort. It shouldn't stop the whole flow. Maybe the user wants the path for something else down the line, i.e., it's going to be used by a later step. Let's not be presumptuous.

donn (changes requested):

kareefardi:
Makes sense. How about this: if a glob doesn't match, keep the `*`, warn the user and continue?

kareefardi:
@donn I found out that there is no current mechanism for tcl.py (and others) to report warnings and I thought that there isn't much added value for a warning. This will simply only resolve the ref in case of failed glob. I believe other parts in the flow can capture "wrong" file path

donn (approved):

donn (approved):
lgtm

Issue #1629, opened 2023-01-17
Set default value for global_verbose_level


hakan-demirli:
### Description

If you follow the Interactive mode [documentation](https://github.com/The-OpenROAD-Project/OpenLane/blob/cbb9e6b02aff0e064e65bd6651157693576de3d8/docs/source/reference/interactive_mode.md) and run any command, a blocking warning/error pops up. ```can't read "global_verbose_level": no such variable```.
Steps to produce:

- Install OpenLane
- ```make mount```
- ```./flow.tcl -design inverter -interactive```
- ```run_synthesis```
Solution is to run ```set_verbose 1``` command to set the mentioned variable. There is only one mention of this variable in #1525 but the main problem of that user was different and no one mentioned set_verbose.

### Proposal

The variable global_verbose_level should have a default value. Maybe also a warning at the beginning of the interactive flow about the default verbosity level. But, the existing logs are enough in my opinion.

PR #1633, opened 2023-01-22
Set default global_verbose_level to 0 in global scope

kareefardi:
Fixes https://github.com/The-OpenROAD-Project/OpenLane/issues/1629


donn (changes requested):
Not a chance. Environment variables ended up getting saved to the `config.tcl` files and confusing people. We fix the problem properly by calling `set_verbose` somewhere appropriate.

Issue #1627, opened 2023-01-16
LEC fails with yosys error


antonblanchard:
### Description

Commit 86615f8def77 ("Yosys Script Enhancements (#1446)") broke LEC. When building DFFRAM I see:

```
ERROR: Module `\sky130_fd_sc_hd__dlclkp_1' referenced in module `\equiv' in cell `\SLICE[3].RAM8.WORD[7].W.BYTE[0].B.genblk1.CG_gate' is not part of the design.
```

### Expected Behavior

DFFRAM builds with CONFIG_LEC enabled

### Environment report

```shell
Distribution: fedora 37
Python: v3.11.1 (OK)
Container Engine: podman v4.3.1 (UNSUPPORTED)
OpenLane Git Version: 03bbd57475d8e2de72b015a8dc5870b651844f6d
pip: INSTALLED
python-venv: INSTALLED
---
PDK Version Verification Status: OK
---
Git Log (Last 3 Commits)

03bbd57 2023-01-16T23:29:39+02:00 Add a check for CURRENT_ODB to or_gui (#1625) - Matt Liberty -  (HEAD -> master, origin/master, origin/HEAD, fix-LEC)
549d7c6 2023-01-16T23:20:13+02:00 [BOT] Update PDK (#1609) - Openlane Bot -  ()
d4b42bd 2023-01-16T03:28:25+02:00 Fix loading design configs (#1623) - Kareem Farid -  ()
---
Git Remotes

origin	https://github.com/The-OpenROAD-Project/OpenLane.git (fetch)
origin	https://github.com/The-OpenROAD-Project/OpenLane.git (push)
```

### Reproduction material

Test case: [issue_reproducible.tar.gz](https://github.com/The-OpenROAD-Project/OpenLane/files/10429646/issue_reproducible.tar.gz)

### Relevant log output

```shell
ERROR: Module `\sky130_fd_sc_hd__dlclkp_1' referenced in module `\equiv' in cell `\SLICE[3].RAM8.WORD[7].W.BYTE[0].B.genblk1.CG_gate' is not part of the design.
```

donn:
Should be fixed now, I think?

kareefardi:
[...]

PR #1641, opened 2023-01-29
Revert "Fix LEC (#1628)"

kareefardi:
This reverts commit 0b94d33d942814fceeab05ec4881c90b5bc386ac. See comment https://github.com/The-OpenROAD-Project/OpenLane/issues/1627#issuecomment-1403476716.


donn:
the year is 2028, the openlane revision history is just "fix lec" "revert fix lec" ad nauseum

donn (commented):
Enough is enough IMO- we either fix it or not. This PR, like the status quo, doesn't work, so there's no point in merging it.

donn (approved):

PR #1647, opened 2023-02-02
Fix wrong file extensions and regular expressions

yusfux:
due to the wrong extensions on "generate_reports.py" and "report.py", we were unable to see the power dist. outputs in the "metrics.csv" and lvs summary in the "manufacturability.rpt".


yusfux:
just realized that even though it fixes the problem with the number of insterted diodes, critical path, and manufact. report, it still cannot give any output about power reports and I believe this is because we do not have any power report with "Slowest/Typical/Fastest Corner" in "rcx_sta.power.rpt" or any file so the if statement in number 364 is kinda broken and it causes the regex in the line 372 to never execute.

donn (approved):
LGTM - thank you

PR #1652, opened 2023-02-06
Fix IR Drop

donn:
+ Add variable `VSRC_LOC_FILE` for PSM voltage source location files
~ Add missing `global_connect` call in `set_global_connections`
~ Overhaul irdrop script, add missing `read_spef`


donn:
Waiting for a loc file for caravel upw from @mo-hosni.

PR #1653, opened 2023-02-06
Added link to tutorials for Spanish learners

erickcb:
I added a link to the tutorials I developed for my class in Costa Rica. These tutorials are written completely in Spanish

Signed-off-by: Erick Carvajal <erick.carvajal.barboza@gmail.com>

### DIFF
--- a/docs/source/additional_material.md
+++ b/docs/source/additional_material.md
@@ -17,6 +17,7 @@ Additionally, we are also going to link to academic publications about OpenLane 
 - [OpenMPW入門 改訂版, VLSI.jp](https://vlsi.jp/OpenMPW.html)
 - [RgGen ✕ OpenMPW Walkthrough](https://vlsi.jp/RgGenxOpenMPW_eng.html#rggen--openmpw-walkthrough)
 - [RgGen ✕ OpenMPWでLSIを焼こう！](https://vlsi.jp/RgGenxOpenMPW.html)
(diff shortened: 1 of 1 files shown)


donn (approved):
I've gone ahead and separated the tutorials by language: but other than that, looks good! Thank you!

kareefardi:
Hey @erickcb Thank you for your contribution. I have some comments about the installation tutorial. I noticed a mixed usage of `make <command>` and `sudo make <command>` which  likely will cause issues. In most cases `sudo` is not needed for the make commands as adding the user to docker group via `sudo usermod -aG docker ${USER}` should be sufficient. Also I am wondering why were the following steps needed (specifically git config and chmod: 
```
    >$ git clone https://github.com/The-OpenROAD-Project/OpenLane.git
    >$ git config --global --add safe.directory 
    >$ cd 
    >$ sudo make
    >$ sudo make test
    >$ sudo chmod 0775 -R 
 ```

erickcb:
Good day Kareem,

Thank you for your comments. In my experience during the course, some
students miss adding the user to the docker group, so I left the "sudos"
just to be safe that the command would work, I will double check this and
leave it only when necessary.

To be honest, those two commands were in the tutorial I followed when I
wrote this, and I am not entirely sure if they are necessary. I will check
this and adjust the tutorials as well.

PR #1668, opened 2023-02-25
sed: use cross-platform form for -i flags

xobs:
The `-i` argument to sed has differing behaviour under BSD and GNU sed:

  - On BSD it requires an argument, though it can be '' (i.e. `-i ''`).
  - Under GNU sed, it needs to immediately follow the argument (i.e. `-i''`).

Provide a backup file argument and immediately delete it, since this appears to be the only cross-platform solution.

With this, as well as some other patches that we are working to upstream, I have successfully built and synthesized a design on a Mac laptop with an ARM64 core.

### DIFF
--- a/scripts/tcl_commands/all.tcl
+++ b/scripts/tcl_commands/all.tcl
@@ -46,7 +46,8 @@ proc set_netlist {args} {
     set ::env(CURRENT_NETLIST) $netlist
 
     set replace [string map {/ \\/} $::env(CURRENT_NETLIST)]
-    try_catch sed -i -e "s/\\(set ::env(CURRENT_NETLIST)\\).*/\\1 $replace/" "$::env(GLB_CFG_FILE)"
+    try_catch sed -i.bak -e "s/\\(set ::env(CURRENT_NETLIST)\\).*/\\1 $replace/" "$::env(GLB_CFG_FILE)"
+    exec rm -f "$::env(GLB_CFG_FILE).bak"
 
     if { [info exists flags_map(-lec)] && $::env(LEC_ENABLE) && [file exists $previous_netlist] } {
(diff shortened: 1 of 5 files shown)


Jenkinsfile
```diff
@@ -80,7 +80,8 @@ pipeline {
                                 }
                                 stage("${DESIGN} - Update OpenROAD commit") {
                                     script { OLD_SHA  = sh (returnStdout: true, script: "python3 dependencies/tool.py -f commit openroad_app").trim(); }
-                                    sh "sed -i s/${OLD_SHA}/${NEW_SHA}/ ./dependencies/tool_metadata.yml"
```
donn, line 83: `Jenkinsfile` only runs in a Linux-based CI environment, where GNU sed is used

docker/cvc_rv/Dockerfile
```diff
@@ -25,7 +25,8 @@ RUN curl -L ${CVC_RV_REPO}/tarball/${CVC_RV_COMMIT} | tar -xzC . --strip-compone
     autoreconf -i && \
     autoconf && \
     ./configure --disable-nls --prefix=/build && \
-    sed -i 's/api.parser.class/parser_class_name/' src/cdlParser.yy && \
```
donn, line 28: Same goes for all the Dockerfiles and build_klayout.sh.

donn (changes requested):
I think I tried this before and it broke for some reason- I'm not opposed to it necessarily in functional OpenLane code but I don't see merit to propagating this change to Docker/CI code, especially that the latter cannot be tested until there's a tool upgrade.

xobs:
Sounds good. I removed all changes from the docker scripts -- now only TCL scripts are changed.

donn (approved):

PR #1670, opened 2023-02-27
Add repair_design for global routing

mo-hosni:
### **Issue:**
The variables for post-routing max wire_length/slew/cap 

- GLB_RESIZER_MAX_WIRE_LENGTH
- GLB_RESIZER_MAX_SLEW_MARGIN
- GLB_RESIZER_MAX_CAP_MARGIN

have default values, but the repair is not done anywhere.

### **Fix**
Added repair_design that uses the previous variables after global routing. This reduces the maximum transition violations drastically in any large design.
Disable routing optimizations for `caravel_upw` design.


antonblanchard:
I did something similar in https://github.com/The-OpenROAD-Project/OpenLane/pull/1024 which we then removed in https://github.com/The-OpenROAD-Project/OpenLane/pull/1033. At the time I wasn't confident we could call `repair_design` later on in the flow without it causing other issues. 

I wonder if repair_timing should take `-max_wire_length` `-slew_margin` and `-cap_margin` options and make an attempt to fix any issues without undoing previous design and timing fixes. Either that, or add an incremental option to `repair_design` that doesn't undo previous fixes. FYI @maliberty

mo-hosni:
@antonblanchard 
> I did something similar in #1024 which we then removed in #1033. At the time I wasn't confident we could call `repair_design` later on in the flow without it causing other issues.

Global routing timing optimization is done after the `max_wire_length`, `slew`, and `cap` repair. So, I don't think there will be an issue in timing. Moreover, I used this on 3 different designs, one of them is a flattened version of caravel and there was no issue in timing. 

> 
> I wonder if repair_timing should take `-max_wire_length` `-slew_margin` and `-cap_margin` options and make an attempt to fix any issues without undoing previous design and timing fixes. Either that, or add an incremental option to `repair_design` that doesn't undo previous fixes. FYI @maliberty

repair_timing does not take these arguments. Check https://openroad.readthedocs.io/en/latest/main/src/rsz/README.html

antonblanchard:
> repair_timing does not take these arguments

I know that, I think you misread my comment.

kareefardi:
@antonblanchard We added a flag to toggle repair_design. Since @mo-hosni tests saw improvements with it we will have it enabled by default.

kareefardi (approved):

PR #1671, opened 2023-02-27
More strict antenna violations repair

mo-hosni:
### **Issue**
Strategies 3 and 6 antenna repair leaves ~300-1000 antenna violations in designs larger than 1mm x 1mm in sky130.

### **Fix**
Increased `GRT_ANT_ITERS` from 3 to 15.
Added `GRT_ANT_MARGIN` which is the margin to over fix antenna violations in global routing with a default value of 10%.


antonblanchard:
With 15 iterations aren't you going to end up with 150 antenna diodes? Also, see https://github.com/The-OpenROAD-Project/OpenROAD/issues/2207

mo-hosni:
> With 15 iterations aren't you going to end up with 150 antenna diodes? Also, see [The-OpenROAD-Project/OpenROAD#2207](https://github.com/The-OpenROAD-Project/OpenROAD/issues/2207)

Well, the largest number of antenna diodes I saw on a single net using `15` iterations was `20` `sky130_fd_sc_hd__diode_2` cells on a ~`3mm x 5mm` design. I am not sure I tried enough designs to see if there will be an issue, but I always use 15 or more antenna repair iterations.

Anyway, 
assuming that the techlef files are correct, can inserting many diodes be worse than leaving antenna violations?

kareefardi (approved):

PR #1672, opened 2023-02-27
Added PL_WIRELENGTH_COEF variable to global placement.

mo-hosni:
### **Added PL_WIRELENGTH_COEF**

`PL_WIRELENGTH_COEF` is the initial wirelength coefficient that is used in global placement. The default value is 0.25. Decreasing `PL_WIRELENGTH_COEF` will modify the initial placement of the standard cells to reduce the wirelengths.


kareefardi (changes requested):
Please document the new variable

kareefardi (approved):

Issue #1599, opened 2023-01-01
FP_PDN_MACRO_HOOKS: No warning/error if instance does not exist


mole99:
### Description

Currently, no warning or error is issued if an instance specified in `FP_PDN_MACRO_HOOKS` does not exist.

For example, with the following entries for [caravel_user_project](https://github.com/efabless/caravel_user_project) the flow will complete successfully, even though `aaaa` does not exist as an instance.

```
"FP_PDN_MACRO_HOOKS": [
    "mprj vccd1 vssd1 vccd1 vssd1,",
    "aaaa vccd1 vssd1 vccd1 vssd1"
],
```

In this obvious case, the wrong entry can be easily spotted. But with subtle typos, it may take a while to find the actual problem.

### Proposal

Check if the modules specified in `FP_PDN_MACRO_HOOKS` actually exist. If not, issue a warning or an error.

d-m-bailey:
@mole99 Your `config.json` file has `"FP_PDN_CHECK_NODES": 0,`.
Doesn't that mean ignore any errors related to power domain connections?

Your final powered netlist contains instances named `\memory[0].sram` but your config file is missing the leading `\`
```
    "FP_PDN_MACRO_HOOKS": [
        "memory[0].sram vccd1 vssd1 vccd1 vssd1,",
        "memory[1].sram vccd1 vssd1 vccd1 vssd1,",
        "memory[2].sram vccd1 vssd1 vccd1 vssd1,",
        "memory[3].sram vccd1 vssd1 vccd1 vssd1,",
        "sram vccd1 vssd1 vccd1 vssd1"
    ],
```

Also `[` and `]` can be tricky, too (they may need to be escaped). There's a translated version of the config.tcl file in `runs/tag` where you can check.

mole99:
Hi @d-m-bailey, thanks for your reply!

I think this answer was meant for #1600, I will answer there.

But related to this issue, I set `"FP_PDN_CHECK_NODES": 1`, which did not generate any warning or error despite the wrong entry in `"FP_PDN_MACRO_HOOKS"`.

PR #1675, opened 2023-02-28
Hierarchical Static Timing Analysis

kareefardi:
\+ Add `EXTRA_SPEFS`
\+ Use `VERILOG_FILES_BLACKBOX` in sta. Allow skipping a file by adding a comment `/// sta-blackbox`
\+ Detect blackboxed modules in typical-corner STA during sign off and warn the user about it
\+ Add `run_sta_script` used in multicorner sta
\+ Support sta in `or_issue.py`
\+ Support sta in `run_tcl_script`
\+ Check if regex defined in `FP_PDN_MACRO_HOOKS` has a match. Fixes https://github.com/The-OpenROAD-Project/OpenLane/issues/1599
\~ Wrap `set_global_connections` around a flag since it is not supported by opensta
\~ Run multicorner sta using standalone sta application for hierarchical
sta support
\~ Move `sta_multi_corner.tcl` to `scripts/openroad/sta/multi_corner.tcl`
\~ Don't delete target directory of `-save` for multicorner spef files
\~ Merge multi_corner sta and single corner into one script.
\~ Split `RUN_STANDALONE` embedded inside openroad steps, which calls sta after, to another sta step run after said openroad step
\- Remove `report_design_area` in `multi_corner.tcl` since it is not supported by standalone OpenSTA 
\- Remove `scripts/openroad/sta.tcl`

---
Fixes https://github.com/The-OpenROAD-Project/OpenLane/issues/1714


kareefardi:
@passant5 @mo-hosni can you check this out?

docs/source/reference/configuration.md
```diff
@@ -35,6 +35,7 @@ These variables are optional that can be specified in the design configuration f
 |Variable|Description|
 |-|-|
 | `VERILOG_FILES_BLACKBOX` | Black-boxed, Verilog files where the implementation is ignored. Useful for pre-hardened macros you incorporate into your design, used during synthesis. |
+| `VERILOG_NETLISTS` | Verilog files of additional modules. Currently, they are used in multiconrer sta. Netlists must be powered. <br> (Default: NONE) |
```
donn, line 38: Typo: conrer. Also, how is this different from `VERILOG_FILES_BLACKBOX`?

docs/source/reference/configuration.md
```diff
@@ -71,6 +72,7 @@ These variables are optional that can be specified in the design configuration f
 
 | Variable | Description |
 |-|-|
+| `MODULES_SPEF_FILES` | Specifies min, nom, max spef files for modules(s). Variable should be provided as a json/tcl list or a space delimited tcl string. Note that a module name is provided not an instance name. A module may have multiple instances. Each module must have 3 files for each corner. For example: `module1 min1 nom1 max1 module2 min2 nom2 max2`. <br> (Default: NONE) |
```
donn, line 75: Would it not be more consistent to call this variable `EXTRA_SPEF`?

Also, can you add the note about using nominal multiple times if the other corners do not exist?

donn (changes requested):
mostly okay, just a couple of qs

docs/source/reference/configuration.md
kareefardi: `VERILOG_FILES_BLACKBOX` is passed to yosys. Same files can't be passed to opensta since opensta only parses gatelevel netlists.

docs/source/reference/configuration.md
kareefardi: Sure

docs/source/reference/configuration.md
donn: Should we not unify? Why should a user have to present multiple Verilog views of the same thing?

VERILOG_FILES_BLACKBOX can also be gate-level netlists.

kareefardi: But it can be rtl as well. In caravel we have simple_por which doesn't have a gl netlist. For that we read in the rtl as a blackbox module. @mo-hosni suggested that we might want to use VERILOG_NETLIST for yosys as well and then VERILOG_FILES_BLACKBOX + VERILOG_NETLISTS as read in by yosys to avoid duplicate files in both variables

donn (commented):
Some small comments

donn (approved):
Welp. Another huge PR. Let's give it a shot.

PR #1676, opened 2023-03-04
Fix a typo in `apply_route_obs`

mo-hosni:
Fixed a typo in `apply_route_obs` in scripts/tcl_commands/routing.tcl.


donn (approved):
Thanks for catching that.

PR #1684, opened 2023-03-16
Rework Placement

donn:
+ Add warning when attempting to run Sylvain's script without `DPL_CELL_PADDING` >= 0
+ Add warning when DPL_CELL_PADDING is set to greater than GPL_CELL_PADDING
~ **Use $HOME/.volare as the default PDK location**
~ `PL_TARGET_DENSITY` incorporates value of `GPL_CELL_PADDING` in calculation
~ Updated various config files to handle new default padding options
~ CI bugfixes + support for Volare 0.7.1

[ci ets]

---

Depends on https://github.com/RTimothyEdwards/open_pdks/pull/358.


donn:
@marwaneltoukhy It triples the design area and is only needed for the custom script, which is disabled by default in favor of DIS 3.

marwaneltoukhy (approved):

PR #1698, opened 2023-03-27
Enhance Synthesis Exit Checks

kareefardi:
~ In `yosys`, call `check -assert` at the correct stage. This requires expanding `yosys` built-in command `synth` to its original sub-commands found in https://github.com/YosysHQ/yosys/blob/master/techlibs/common/synth.cc and inject `check -assert` in the middle of the expanded command. Experimentally, this was the only way to properly capture used wires that do not have a driver.

Additional synthesis checkers guarded with `QUIT_ON_SYNTH_CHECKS`:
\+ Add `check_latches`
\+ Add `check_out_of_bound`
\+ Add `check_resizing_cell_port`

~ Fix a "resizing cell port" error in `APU` design
~ Fix a "used but has no driver" error in `tests/912`

### DIFF
--- a/designs/APU/src/APU.v
+++ b/designs/APU/src/APU.v
@@ -675,8 +675,8 @@ assign odd_or_even = InternalClock;
 
 
 // Generate each channel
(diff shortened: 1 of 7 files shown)


donn (approved):
As in, approve LGTM, dear god github

Issue #1711, opened 2023-04-01
Provide option to increase pin spacing with pin_order.cfg


vijayank88:
### Description

Based on comment https://github.com/The-OpenROAD-Project/OpenLane/issues/1699#issuecomment-1488316946
For IO limited block with fixed pin location through `pin_order.cfg` it is good to add `FP_IO_MIN_DISTANCE` configuration to keep some minimum distance.

![pin_order_issue](https://user-images.githubusercontent.com/85787913/229271645-9b6643ca-83fb-49e6-8894-346886396c72.png)
in above floorplan, the corner has some more space. As existing pin placement not increasing pin distance. It is good to customize pin distance for better pin placement.

So it will be easy to place buffer and routing, otherwise facing congestion issue.

kareefardi:
@vijayank88 what is the exact configuration that you used to take that screenshot?

vijayank88:
Increased die size 1600x3010 and ran the OpenLane flow with user design of fixed pin_order.cfg.

vijayank88:
Yes, I agree. Die size should be increased to place all the pins...

hanm2019:
make be place in multiple layer? Can I decide the pin's layer also in pin_order.cfg? such as
```
#LAYER 1
#N 
clk
rst
bypass

#S
xxxx

#LAYER 2
#N 
...
...

#S
...
...

```

PR #1715, opened 2023-04-04
`io_place.py` Fixes

kareefardi:
Fixes https://github.com/The-OpenROAD-Project/OpenLane/issues/1711
\+ integrate `FP_IO_MIN_DISTANCE`
\~ respect min spacing while placing pins
\~ print on which side requested pins exceed allowed number of pins

others:
\~ tap_decap step generate output to tmp instead of result
\~ gen_pdn step generate output to result instead of tmp because it is the final step in floorplan stage


donn (approved):
LGTM

PR #1723, opened 2023-04-06
`io_place.py` pin separation constraints adjustments

kareefardi:
\~ set min distance in cfg file instead of using `FP_IO_MIN_DISTANCE`
\~ don't quit when min distance < legal min distance and use legal min distance. by definition if min distance is illegal it is "minimum" and we can use a higher value
\~ `min_distance` is not min spacing - adjust legal tracks in `io_place.py` under that assumption


kareefardi:
@donn do you think we should always `puts_warn` that FP_IO_MIN_DISTANCE has no use with io_place.py

donn (approved):
LGTM- good work

donn:
Also, no. If we were to warn people variables for one step wouldn't affect another, we'd be doing this for the next 8 years.

Issue #1725, opened 2023-04-08
LEF file is broken on the latest OpenLane version


hakan-demirli:
### Description

LEF files are broken on the latest OpenLane version. No matter the design following LEF file is generated. This is for the spm. It is generated after ```make test``` command.
```
VERSION 5.7 ;
  NOWIREEXTENSIONATPIN ON ;
  DIVIDERCHAR "/" ;
  BUSBITCHARS "[]" ;
MACRO spm
  CLASS BLOCK ;
  FOREIGN spm ;
  ORIGIN 0.000 0.000 ;
  SIZE 0.005 BY 0.005 ;
END spm
END LIBRARY
```
If I were to create a macro and use the LEF file together with the other generated files like GDS and LIB, flow fails since the area is 0.005x 0.005 um^2.

Tag `2023.03.12` is working without any problems. Area and other fields are correct and I can use the macros generated.

### Expected Behavior

Proper LEF file is expected. 

### Environment report

```shell
Kernel: Linux v5.15.0-69-generic
Distribution: ubuntu 22.04
Python: v3.10.6 (OK)
Container Engine: docker v23.0.3 (OK)
OpenLane Git Version: cb634fd5f670db4a12e36075c642dae54150955c
pip: INSTALLED
python-venv: INSTALLED
---
PDK Version Verification Status: OK
---
Git Log (Last 3 Commits)

cb634fd 2023-04-06T13:41:52+02:00 Fix ci go package install (#1722) - Kareem Farid -  (grafted, HEAD -> master, tag: 2023.04.07, origin/master, origin/HEAD)
---
Git Remotes

origin	ssh://git@github.com/The-OpenROAD-Project/OpenLane (fetch)
origin	ssh://git@github.com/The-OpenROAD-Project/OpenLane (push)
```

### Reproduction material

```bash 
git clone https://github.com/The-OpenROAD-Project/OpenLane
cd OpenLane
make
make test
```

### Relevant log output

```shell
VERSION 5.7 ;
  NOWIREEXTENSIONATPIN ON ;
  DIVIDERCHAR "/" ;
  BUSBITCHARS "[]" ;
MACRO spm
  CLASS BLOCK ;
  FOREIGN spm ;
  ORIGIN 0.000 0.000 ;
  SIZE 0.005 BY 0.005 ;
END spm
END LIBRARY
```

kareefardi:
I can duplicate the issue and will work on it

PR #1728, opened 2023-04-10
Magic LEF write updates

kareefardi:
```
~ use lef/def or gds instead of mag to write a lef through MAGIC_LEF_WRITE_USE_GDS
+ add wrapper.tcl to capture magic errors ported from openlane2
~ print lef write log file path
```
Fixes https://github.com/The-OpenROAD-Project/OpenLane/issues/1725


donn (approved):
LGTM- make sure to port this to OpenLane 2

Issue #1726, opened 2023-04-10
Generate OpenROAD gui script as part of results


xobs:
### Description

OpenROAD has a very powerful GUI that requires some setup in order to get it working. Notably, the GUI must be passed a series of files to take maximum advantage of the information produced during build synthesis.

OpenLane could generate a script file as part of the release process that can be passed directly to openroad.

### Proposal

As an example, I have a script that loads the db, liberty, sdc, and spef files from a project:

```
read_db results/routing/minimax.odb
read_liberty /si/conda/share/pdk/gf180mcuC/libs.ref/gf180mcu_fd_sc_mcu7t5v0/liberty/gf180mcu_fd_sc_mcu7t5v0__tt_025C_5v00.lib
read_sdc results/final/sdc/minimax.sdc
read_spef results/final/spef/minimax.spef
```

Such a script could be generated very easily as part of the final generation process.

maliberty:
Note that flow.tcl has a -gui option already but it needs some work to correctly setup timing.

kareefardi:
Does https://github.com/The-OpenROAD-Project/OpenLane/pull/1729 addresses your concerns?
@xobs you can run `flow.tcl -design spm -tag tag1 -gui` to open up a (partially) finished run for spm with tag `tag1`. The above PR also load spef and sdc.

@maliberty `estimate_parastics` needs a flag. I am not aware of a direct way to find out if global routing has been run so I chose to print a message.

xobs:
That does address it, mostly. I didn't know that `flow.tcl` can be used to inspect the results of a run, in addition to running a flow.

I think the bigger issue is that the `-gui` flag doesn't appear to be documented, which may be why I didn't know it existed: https://openlane.readthedocs.io/en/latest/search.html?q=-gui&check_keywords=yes&area=default

It might be a good idea to document the existance of that option, so people know to look for it.

maliberty:
@kareefardi you could either look at what step in the flow is being loaded and decide based on that or you could look in odb and see if there are any guides present.

Issue #1692, opened 2023-03-22
load timing data in gui.tcl


maliberty:
### Description

When you run the flow with -gui the layout is loaded with the SDC & Liberty but no parasitics are loaded (or estimated).  

### Proposal

Load the spef parasitics if available or otherwise estimate some.  This will enable timing to be meaningfully used.

(I don't have time to work on this but hope someone will pick it up)

donn:
Fixed in #1729

PR #1729, opened 2023-04-10
read spef and sdc in -gui

kareefardi:
Fixes https://github.com/The-OpenROAD-Project/OpenLane/issues/1726 and https://github.com/The-OpenROAD-Project/OpenLane/issues/1692

### DIFF
--- a/scripts/openroad/gui.tcl
+++ b/scripts/openroad/gui.tcl
@@ -13,3 +13,13 @@
 # limitations under the License.
 source $::env(SCRIPTS_DIR)/openroad/common/io.tcl
 read
+if { [info exists ::env(CURRENT_SPEF)] } {
+    puts "Reading spef $::env(CURRENT_SPEF)"
+    read_spefs
+} else {
+    puts "WARNING no spef found for the current run. Please either run estimate_parasitics -placement or estimate_parasitics -global_routing"
+}
+if { [info exists ::env(CURRENT_SDC)] } {
+    puts "Reading sdc $::env(CURRENT_SDC)"
+    read_sdc $::env(CURRENT_SDC)
+}


scripts/openroad/gui.tcl
```diff
@@ -12,4 +12,14 @@
...
-read
\ No newline at end of file
+read
+if { [info exists ::env(CURRENT_SPEF)] } {
+    puts "Reading spef $::env(CURRENT_SPEF)"
+    read_spefs
+} else {
+    puts "WARNING no spef found for the current run. Please either run estimate_parasitics -placement or estimate_parasitics -global_routing"
+}
+if { [info exists ::env(CURRENT_SDC)] } {
```
donn, line 22: This is already in `read`.

scripts/openroad/gui.tcl
```diff
@@ -12,4 +12,14 @@
...
 # limitations under the License.
 source $::env(SCRIPTS_DIR)/openroad/common/io.tcl
-read
\ No newline at end of file
+read
+if { [info exists ::env(CURRENT_SPEF)] } {
+    puts "Reading spef $::env(CURRENT_SPEF)"
+    read_spefs
+} else {
+    puts "WARNING no spef found for the current run. Please either run estimate_parasitics -placement or estimate_parasitics -global_routing"
```
donn, line 20: Not really in-line with how we print warnings and such in OpenLane...

donn (changes requested):
I'll fix these myself. One sec.

donn (approved):

scripts/openroad/gui.tcl
kareefardi: I don't want to source utils.tcl so I guess I will duplicate puts_warn

PR #1731, opened 2023-04-10
Fix GDS write fail by klayout

hpretl:
If the `os.path.abspath` is added to the `design_name`, then this causes a fail of the GDS write by `klayout` (see Slack discussion with @donn).


donn (approved):
LGTM- will merge. Thank you.

Issue #1735, opened 2023-04-10
`check_synth_misc` doesn't apply when `SYNTH_ELABORATE_ONLY` is set to 1 


kareefardi:
### Description

Title.

### Expected Behavior

`elaborate.tcl` outputs a check file we should: (1) validate the checks file and adjust `elaborate.tcl` if needed (2) guarantee that check_synth_misc works when `SYNTH_ELABORATE_ONLY` 

### Environment report

```shell
N/A
```

### Reproduction material

N/A

### Relevant log output

```shell
N/A
```

PR #1736, opened 2023-04-10
Run full synth checks when synth elaborate is on

kareefardi:
```
+ set pre_synth_report depending on SYNTH_ELABORATE_ONLY
~ adjust _EXT variables in synth.tcl
~ better log message for /// sta-blackbox
- remove multiple drivers in caravel_upw and
keep the power connections only as we are mainly interested in them
```
---
Fixes https://github.com/The-OpenROAD-Project/OpenLane/issues/1735


donn:
caravel

donn (approved):
LGTM

PR #1743, opened 2023-04-11
Don't unset env variables after executing a step

kareefardi:
To avoid causing issues when attempting to run `or_issue.py` outside of the flow (in case it didn't crash)

### DIFF
--- a/scripts/tcl_commands/magic.tcl
+++ b/scripts/tcl_commands/magic.tcl
@@ -153,9 +153,6 @@ proc run_magic_spice_export {args} {
     run_magic_script $::env(SCRIPTS_DIR)/magic/extract_spice.tcl\
         -indexed_log $log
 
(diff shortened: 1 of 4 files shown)


donn:
No. You can already run `or_issue` without a crash by either adding a nonzero exit to the script or simply by exporting `CREATE_REPRODUCIBLE_FROM_SCRIPT=script.tcl` before running the flow. I see no reason to accept the chaos this may potentially cause.

kareefardi:
> No. You can already run `or_issue` without a crash by either adding a nonzero exit to the script or simply by exporting `CREATE_REPRODUCIBLE_FROM_SCRIPT=script.tcl` before running the flow. I see no reason to accept the chaos this may potentially cause.

This wouldn't work with finished runs. Why would this cause chaos ?

donn:
Future steps may be sensitive to these temporary variables.

PR #1760, opened 2023-04-17
Integrate `Verilator`

kareefardi:
```
+ Add verilator check before synthesis
+ Add QUIT_ON_VERILATOR_ERRORS
+ Add QUIT_ON_VERILATOR_WARNINGS
+ Add VERILATOR_RELATIVE_INCLUDES
+ Only load verilog models for selected PDKs and warn the user about unsupported PDK
```


configuration/synthesis.tcl
```diff
@@ -33,3 +33,7 @@ set ::env(IO_PCT) 0.2
 set ::env(SYNTH_EXTRA_MAPPING_FILE) ""
 
 set ::env(BASE_SDC_FILE) $::env(SCRIPTS_DIR)/base.sdc
+
+set ::env(QUIT_ON_VERILATOR_WARNINGS) 0
+set ::env(QUIT_ON_VERILATOR_ERRORS) 1
+set ::env(VERILATOR_VERIFIED_PDKS) "sky130A sky130B"
```
donn, line 39: I don't think this should be a variable, rather, this should be built into the step itself.

flow.tcl
```diff
@@ -177,6 +177,7 @@ proc run_non_interactive_mode {args} {
     set ANTENNACHECK_ENABLED [expr ![info exists flags_map(-no_antennacheck)] ]
 
     set steps [dict create \
+        "verilator_lint" "run_verilator" \
```
donn, line 180: Why not `run_verilator_step` and have a `RUN_VERILATOR` variable, to match some of the other steps?

donn (commented):
Some notes

kareefardi:
@donn Fixed notes

Another thing: I noticed that arm64v8 build takes around three hours. Can we do something there - aside from building and pushing locally ?

donn (commented):
LGTM, will merge after CI

donn:
@kareefardi Realistically, only two things can be done here:

* Self-hosted runners, which are a nightmare.
* Cross-compilation. However, I did try that early on but it didn't work out.

For OpenLane 2, Nix has cross-compilation as a first-class citizen, solving that problem, but for OL1 I just decided to eat up the time for using an emulator: https://nix.dev/tutorials/cross-compilation

kareefardi:
CI failing. Checking testcases.

kareefardi:
TODO: Add option to load verilog models of PDK

donn (approved):

PR #1761, opened 2023-04-17
Remove patch from OpenROAD Docker build

kareefardi:


donn (approved):

kareefardi:
@hpretl Patch is not needed now

hpretl:
@kareefardi Thanks for the ping, I also removed the patch already.

PR #1766, opened 2023-04-18
Reformat `io.tcl`

donn:
~ Use ellipsis instead of three periods
~ Add quotes to file paths so they can be navigated to in IDEs and text editors


kareefardi:
LGTM. Waiting for CI

PR #1775, opened 2023-04-27
Fix `run_path` not passed to `ConfigHandler`

dlmiles:
Since 2023.03.28 when this line of code was added when -run_path is specified on the command line and -run_path is different from the -design path, reporting would error trying to load config.tcl from the wrong location.


dlmiles:
See log documentation at: https://github.com/The-OpenROAD-Project/OpenLane/issues/1777

kareefardi:
Thanks!

PR #1776, opened 2023-04-27
synthesis.tcl: run_verilator --lint-only option changes

dlmiles:
Rationale

--no-timing make it not error on timing statements:
  assign #1 out = !in;

+define+SYNTHESIS=1
  as this pass is on behalf of the synthesis phase so should not include
  verilog which is not for synthesis.  Such as blackboxed primitives.

+define+SYNTHESIS_VERILATOR_LINT_ONLY=1
  like the above but provide more explicit control over this specific
  OpenROAD task to ignore.

  Maybe this option should be added to documentation ?

### DIFF
--- a/scripts/tcl_commands/synthesis.tcl
+++ b/scripts/tcl_commands/synthesis.tcl
@@ -271,6 +271,9 @@ proc run_verilator {} {
     lappend arg_list {*}$arg
     try_exec bash -c "verilator \
         --lint-only \
+        --no-timing \
+        +define+SYNTHESIS=1 \
+        +define+SYNTHESIS_VERILATOR_LINT_ONLY=1 \
         -Wall \
         --Wno-DECLFILENAME \
         --top-module $::env(DESIGN_NAME) \


dlmiles:
Can rename 2nd define if you don't like the label use and add to documentation upon review.

Please provide link to the section of documentation it should be put.

kareefardi:
Thanks @dlmiles. Can you elaborate how should one use `SYNTHESIS` and `SYNTHESIS_VERILATOR_LINT_ONLY`

dlmiles:
Because the input Verilog can contain:

``` `ifndef SYNTHESIS```

Around blackbox primitive items, the synthesiser is expected to provide these by another means.
_The purpose of this --line-only (I assume) is a pre-check on behalf of the SYNTHESIS pass (based on its location in the synthesis.tcl file, and the sequence it occurs in the flow.tcl)_

 And indeed it does, this pass already includes the cell files from the PDK directly:

```
%Warning-MODDUP: //work/src/clock_stretch_invert_mux.v:87:8: Duplicate declaration of module: 'sky130_fd_sc_hd__clkbuf_8'
   87 | module sky130_fd_sc_hd__clkbuf_8 (
      |        ^~~~~~~~~~~~~~~~~~~~~~~~~
                 //home/runner/pdk/sky130A/libs.ref/sky130_fd_sc_hd/verilog/sky130_fd_sc_hd.v:23661:8: ... Location of original declaration
23661 | module sky130_fd_sc_hd__clkbuf_8 (
      |        ^~~~~~~~~~~~~~~~~~~~~~~~~
                 ... For warning description see https://verilator.org/warn/MODDUP?v=5.009
                 ... Use "/* verilator lint_off MODDUP */" and lint_on around source to disable this message.


%Warning-MODDUP: //work/src/glitch_free_clock_mux.v:90:8: Duplicate declaration of module: 'sky130_fd_sc_hd__clkbuf_4'
   90 | module sky130_fd_sc_hd__clkbuf_4 (
      |        ^~~~~~~~~~~~~~~~~~~~~~~~~
                 //home/runner/pdk/sky130A/libs.ref/sky130_fd_sc_hd/verilog/sky130_fd_sc_hd.v:23561:8: ... Location of original declaration
23561 | module sky130_fd_sc_hd__clkbuf_4 (
      |        ^~~~~~~~~~~~~~~~~~~~~~~~~
```


My verilog code looks like:


clock_stretch_invert_mux.v:
```
93335f48 (Darryl L. Miles 2023-04-24 15:14:41 +0100 73)     sky130_fd_sc_hd__clkbuf_8 myclkbuf8(
93335f48 (Darryl L. Miles 2023-04-24 15:14:41 +0100 74)         .A(mux_or[0] | mux_or[1] | mux_or[2]),
93335f48 (Darryl L. Miles 2023-04-24 15:14:41 +0100 75)         .X(clock_out)
93335f48 (Darryl L. Miles 2023-04-24 15:14:41 +0100 76)     );
93335f48 (Darryl L. Miles 2023-04-24 15:14:41 +0100 77) 
455ae660 (Darryl L. Miles 2023-04-23 11:41:31 +0100 78)     // Glitch free clock mux output final 3-input OR
93335f48 (Darryl L. Miles 2023-04-24 15:14:41 +0100 79)     //assign clock_out = mux_or[0] | mux_or[1] | mux_or[2];     // need for iverilog
455ae660 (Darryl L. Miles 2023-04-23 11:41:31 +0100 80)     //assign clock_out = |mux_or[2:0];          // reduction operator
455ae660 (Darryl L. Miles 2023-04-23 11:41:31 +0100 81)     //assign clock_out = |mux_or;               // reduction operator
455ae660 (Darryl L. Miles 2023-04-23 11:41:31 +0100 82) 
455ae660 (Darryl L. Miles 2023-04-23 11:41:31 +0100 83) endmodule
93335f48 (Darryl L. Miles 2023-04-24 15:14:41 +0100 84) 
93335f48 (Darryl L. Miles 2023-04-24 15:14:41 +0100 85) `ifndef SYNTHESIS
93335f48 (Darryl L. Miles 2023-04-24 15:14:41 +0100 86) `ifndef GL_TEST
93335f48 (Darryl L. Miles 2023-04-24 15:14:41 +0100 87) module sky130_fd_sc_hd__clkbuf_8 (
93335f48 (Darryl L. Miles 2023-04-24 15:14:41 +0100 88)     output X,
93335f48 (Darryl L. Miles 2023-04-24 15:14:41 +0100 89)     input A
93335f48 (Darryl L. Miles 2023-04-24 15:14:41 +0100 90) );
93335f48 (Darryl L. Miles 2023-04-24 15:14:41 +0100 91) 
93335f48 (Darryl L. Miles 2023-04-24 15:14:41 +0100 92)     assign X = A;
93335f48 (Darryl L. Miles 2023-04-24 15:14:41 +0100 93) 
93335f48 (Darryl L. Miles 2023-04-24 15:14:41 +0100 94) endmodule
93335f48 (Darryl L. Miles 2023-04-24 15:14:41 +0100 95) `endif
93335f48 (Darryl L. Miles 2023-04-24 15:14:41 +0100 96) `endif
```

glitch_free_clock_mux.v
```
93335f48 (Darryl L. Miles 2023-04-24 15:14:41 +0100 79)     sky130_fd_sc_hd__clkbuf_4 myclkbuf4(
93335f48 (Darryl L. Miles 2023-04-24 15:14:41 +0100 80)         .A(and02 | and12),
93335f48 (Darryl L. Miles 2023-04-24 15:14:41 +0100 81)         .X(clk_out)
93335f48 (Darryl L. Miles 2023-04-24 15:14:41 +0100 82)     );
93335f48 (Darryl L. Miles 2023-04-24 15:14:41 +0100 83) 
93335f48 (Darryl L. Miles 2023-04-24 15:14:41 +0100 84)     //assign clk_out = and02 | and12;
93335f48 (Darryl L. Miles 2023-04-24 15:14:41 +0100 85) 
93335f48 (Darryl L. Miles 2023-04-24 15:14:41 +0100 86) endmodule
93335f48 (Darryl L. Miles 2023-04-24 15:14:41 +0100 87) 
93335f48 (Darryl L. Miles 2023-04-24 15:14:41 +0100 88) `ifndef SYNTHESIS
93335f48 (Darryl L. Miles 2023-04-24 15:14:41 +0100 89) `ifndef GL_TEST
93335f48 (Darryl L. Miles 2023-04-24 15:14:41 +0100 90) module sky130_fd_sc_hd__clkbuf_4 (
93335f48 (Darryl L. Miles 2023-04-24 15:14:41 +0100 91)     output X,
93335f48 (Darryl L. Miles 2023-04-24 15:14:41 +0100 92)     input A
93335f48 (Darryl L. Miles 2023-04-24 15:14:41 +0100 93) );
93335f48 (Darryl L. Miles 2023-04-24 15:14:41 +0100 94) 
93335f48 (Darryl L. Miles 2023-04-24 15:14:41 +0100 95)     assign X = A;
104e59cf (Darryl L. Miles 2023-04-24 15:14:41 +0100 96) 
d5af826c (Darryl L. Miles 2023-04-24 15:14:41 +0100 97) endmodule
93335f48 (Darryl L. Miles 2023-04-24 15:14:41 +0100 98) `endif
93335f48 (Darryl L. Miles 2023-04-24 15:14:41 +0100 99) `endif
```


The purpose of the ``` `ifndef SYNTHESIS_VERILATOR_LINT_ONLY```  was to provide an ifdef pre-processor directive that can be used specifically target this specific pass (veriltator --lint-only on behalf on synthesis) inside OpenLANE.  This would need to be a distinct mechanism and would allow someone to fine tune exactly which lines of Verilog Verilator gets to see in this pass.

I have no direct use case myself for this today, but I can envisage someone wanting the additional checking verilator provides but having a project that may fail/error due to differences between verilator/yosys in Verilog feature support causing the pass to error for verilator for not yosys.

The alternative is to use `RUN_VERILATOR=0` the big red switch on the pass.

**Comment by kareefardi** [comment]
For duplicate declaration, this is due to including the PDK verilog models with Verilator to avoid the need for having user-defined definitions for PDK models. This is similar to the approach of `SYNTH_READ_BLACKBOX_LIB` that allows usage of PDK modules without having to manually define blackbox models. Perhaps that's not the best practice for Verilator ?

If I understand correctly, a more general approach to what you are saying is having a list of `DEFINES` where each should be passed to Verilator. Hence, the way to proceed here for this PR, is to create two variables `VERILATOR_TIMING` that passes `--timing` or `--no-timing` and another variable represented as a list for defines, for example `VERILATOR_DEFINES`. Both should be documented with variables

**Comment by dlmiles** [comment]
This is a LINT ONLY check ?  It is not doing any real work and is supplementary to the flow.tcl end goal?  That is to generate GDS. 

Can you confirm the rationale for adding run_verilator (Verilator --lint-only) is get access to Verilator warnings and error messages to improve code quality ?  This is certainly a nice to have.  The real Verilog parser in the flow.tcl is yosys.

How does VERILATOR_TIMING work for LINT ONLY usage ?  It is not going to simulate, only check the syntax of the Verilog.  Maybe this is really an indication of a Verilator bug/oversight, it does look like the timing command line arguments have been recently added to Verilator but maybe the --lint-only use case not thought out (by the Verilator team).  Maybe Verilator should not be erroring ?  So maybe this is a temporary thing.


 I don't see any issue with the primitives being provided as they already are.  That is the correct thing to do.  I don't consider that aspect a bug (for the purpose of this report).

A `VERILATOR_DEFINES` mechanism and/or a `VERILATOR_ARGUMENTS` mechanism sounds like it will allow more control over the options, so that also sounds good to me.  Maybe the default values should be the ones I have provided.
Maybe the `--no-timing` would be better added to the `VERILATOR_ARGUMENTS` in the config.


The only purpose of my primitives (in my verilog code) is to allow pre-synth simulation to work with the same code base (when sky130 primitives have been instantiated directly).  My version of the primitive MUST NOT BE USED for GDL flow, they are fake and maybe incorrect.  This TCL script is performing work on behalf of the synthesis process/pass so is expected to provide the real versions from the PDK.  Just like it does already.

So it would seem correct to not include any lines of verilog that have been marked "not for synthesis" and the industry standard way of doing that is to use `ifndef SYNTHESIS`.  Adding `SYNTHESIS=1 SYNTHESIS_VERILATOR_LINT_ONLY=1` to the `VERILATOR_DEFINES` which are then converted into their `+define+NAME+VALUE` argument form would certainly provide a mechanism to achieve the same goals.

To summarise:

```
VERILATOR_ARGUMENTS="--no-timing"
VERILATOR_DEFINES="SYNTHESIS=1 SYNTHESIS_VERILATOR_LINT_ONLY=1"
```

This looks a good solution to me.  I think VERILATOR_TIMING is the wrong solution here, for the --lint-only use case.

While OpenLANE is shipping a Verilator version that both supports --no-timing and causes an error for a --lint-only usage if timing information exists in the verilog source code.  Then it should probably be the default setting for config.tcl.

kareefardi:
I agree that `VERILATOR_TIMING` is not a good name, however using `VERILATOR_ARGUMENTS` is pretty unconventional for OpenLane. We tend to expose options using seperate variables. We can come up with a better name than `VERILATOR_TIMING` rather than aggregate multiple flags in a single variable.

dlmiles:
I am not aware of OpenLANE conventions and how they might impact a resolution.

Solution 1: Having a generic general purpose ability to specify any command line argument `VERILATOR_ARGUMENTS`, feels like it has the right amount of effort to implement, low maintenance in the future, addresses future problems in a way that doesn't require a new OpenLANE release to fix.

Solution 2: Having unique environment variable key/value pairs for each and every option`VERILATOR_TIMING` of every thing seems like a high effort (to implement each option), high maintenance (every unique tool), needs a new OpenLANE release to fix (because the specific option for the specific thing is not implemented yet), you will always be playing whack-a-mole to upstream tooling.

Maybe moving to a `config.json` (and away from `config.tcl` and environment like variables) where an entire hierarchical section can be devoted to defining the command line options for verilator would be more inline with solution 1 than solution 2, does that still make it unconventional ?

.. 

What action is expected of me on this PR ?  If someone who is more knowledgeable on OpenLANE conventions and methods to implement is able to resolve that is great.  I do currently have a script that patches OpenLANE as necessary for my use case with this commit (and that works for me), but I suspect at some point others will also report the same issue with the addition with the use of `verilator --lint-only` in the flow.tcl.  For which their resolution maybe `RUN_VERILATOR=0` which voids all benefits of including it :( for them.

Maybe I have spare capacity to look at a revised solution/patch, if I fully understand exactly what solution needs to look like and agree it is a benefit to how I would like a future OpenLANE to look/work/behave.

kareefardi:
@dlmiles You have valid points. We are also trying to move away from "\*VERILATOR\*" variables (see https://github.com/The-OpenROAD-Project/OpenLane/issues/1790). I think we have quite discussed the intention behind the PR. As Verilator integration is fairly new, let me discuss this with the team and see what is the proper way to move forward.

dlmiles:
Feel free to close this PR (unmerged) if the sentiments are logged and being worked on.  Thanks

Issue #1780, opened 2023-05-01
Escaped backslash mis-handled in json config files


kareefardi:
### Description
A variable defined as `\\foo` in a json config file is represented in `config_in.tcl` and as `\\foo`

### Expected Behavior

Value in `config_in.tcl` should be `\foo` since variables stored in config_in.tcl are wrapped in curly braces.

### Environment report

```shell
Kernel: Linux v6.1.20-060120-generic
Distribution: ubuntu 22.04
Python: v3.10.6 (OK)
Container Engine: docker v20.10.22 (OK)
OpenLane Git Version: 0d205c619229eb3ba227e4a7fc042d344b8b6d07
pip: INSTALLED
python-venv: INSTALLED
---
PDK Version Verification Status: OK
---
Git Log (Last 3 Commits)

0d205c61 2023-04-27T13:21:06+02:00 Rename `PDN_CFG` to `FP_PDN_CFG` (#1773) - Kareem Farid -  (HEAD -> master, origin/master, origin/HEAD)
a9419c0c 2023-04-27T12:23:22+02:00 [BOT] Update magic (#1758) - Openlane Bot -  ()
466eb0ba 2023-04-26T13:14:14+02:00 [BOT] Update openroad_app (#1695) - Openlane Bot -  (tag: 2023.04.27)
---
Git Remotes

fork	git@github.com:kareefardi/OpenLane.git (fetch)
fork	git@github.com:kareefardi/OpenLane.git (push)
origin	git@github.com:The-OpenROAD-Project/OpenLane.git (fetch)
origin	git@github.com:The-OpenROAD-Project/OpenLane.git (push)
```

### Reproduction material

N/A

### Relevant log output

```shell
N/A
```

PR #1781, opened 2023-05-01
Do not escape double backslash when translating from json to tcl

kareefardi:
Fixes https://github.com/The-OpenROAD-Project/OpenLane/issues/1780


donn:
Handled in #1782

PR #1803, opened 2023-05-14
get_tag.py: Don't encode illegal docker tag when local building

dlmiles:
I was getting an encoding of "heads/2023.04.19-dev-amd64" and it seems the slash is not allowed to be there.

### DIFF
--- a/dependencies/get_tag.py
+++ b/dependencies/get_tag.py
@@ -23,6 +23,11 @@ class NoGitException(Exception):
     pass
 
 
+# Slashes are not allowed in the container tag
+def canon(ref: str) -> str:
+    return ref.replace("/", "-")
+
+
 def get_tag() -> str:
     try:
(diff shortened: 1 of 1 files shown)


donn:
Hi Darryl, thank you for your contribution- I'm more miffed as to why `git rev-parse HEAD` is returning an invalid branch name..? The mysteries of git.

I believe the solution would be to split by "/" and get the last element, not to attempt to escape "/" somehow.

kareefardi:
> Hi Darryl, thank you for your contribution- I'm more miffed as to why `git rev-parse HEAD` is returning an invalid branch name..? The mysteries of git.
> 
> I believe the solution would be to split by "/" and get the last element, not to attempt to escape "/" somehow.

You can also create a new branch that has `/` in it.

kareefardi (approved):

PR #1836, opened 2023-06-06
Add --timing to verilator options

mguthaus:
This allows behavioral SRAM macros to implement hold and setup delays. Otherwise, Verilator would complain and fail that this (or --no-timing) needs to be specified.


donn:
Folded into #1837

PR #1840, opened 2023-06-07
Deprecate  `-gui` and continuing the flow with an existing tag

kareefardi:
`-gui` is superseded by `gui.py`. `or_gui` should still work for interactive scripts. 
Dropping support for continuing existing runs. Users need to use `-overwrite` or provide a different tag.

### DIFF
--- a/flow.tcl
+++ b/flow.tcl
@@ -157,7 +157,7 @@ proc run_non_interactive_mode {args} {
         {-save_path optional}
         {-override_env optional}
     }
-    set flags {-save -run_hooks -no_lvs -no_drc -no_antennacheck -gui}
+    set flags {-save -run_hooks -no_lvs -no_drc -no_antennacheck -gui -overwrite}
     parse_key_args "run_non_interactive_mode" args arg_values $options flags_map $flags -no_consume
 
     if { [info exists arg_values(-from) ] || [info exists arg_values(-to)] } {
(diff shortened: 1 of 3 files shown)


mguthaus:
Dropping -overwrite requires that a user manually delete a run directory every time they want to re-run with a tag. otherwise, the run will continue starting with the last step of the previous run. I believe this wll result in quite a bit of user confusion.

Suggestion: 
Either keep -overwrite or fail if a run directory of a given tag already exists.

mguthaus:
Defaulting to overwrite without warning could be catastrophic, and is not recommended.

kareefardi:
I wanted to drop support for rerunning existing tag, but I guess I can do that while requiring `-overwrite` to re-use a tag.

flow.tcl
```diff
@@ -165,13 +165,14 @@ proc run_non_interactive_mode {args} {
         exit -1
     }
 
-    prep {*}$args
-    # signal trap SIGINT save_state;
-
     if { [info exists flags_map(-gui)] } {
-        or_gui
+        puts_warn "Flag -gui is now deprecated. Refer to https://openlane.readthedocs.io/en/latest/reference/gui.html to view files graphically."
         return
```
donn, line 170: Exit should be non-zero.

donn (changes requested):

donn (approved):

PR #1908, opened 2023-07-18
remove `unset_propagated_clock` 

passant5:
removed `unset_propagated_clock` from:
1- `/scripts/openroad/floorplan.tcl`
2- `/scripts/openroad/resizer.tcl`
3- `/scripts/openroad/sta/multi_corner.tcl`

To avoid using a generated SDC that doesn't include `set_propagated_clock` in signoff STA. Further review is in the works to determine if the removed commands are required.


donn (approved):
LGTM - waiting on CI

Issue #1925, opened 2023-08-02
Update to python API


vvbandeira:
### Description

OpenROAD made recent changes to the `odb` python API.
See https://github.com/The-OpenROAD-Project/OpenROAD/pull/3728

### Proposal

Update `odb` python API calls. For example, at https://github.com/The-OpenROAD-Project/OpenLane/blob/master/scripts/odbpy/reader.py#L37 use instead:

```python
odb.read_def(self.db.getTech(), def_in)
```

PR #1926, opened 2023-08-02
OpenROAD python `read_def` api updates

kareefardi:
bump openroad version
update openroad python odb read_def api usage to pass tech instead of odb object based on https://github.com/The-OpenROAD-Project/OpenLane/issues/1925

---
Fixes https://github.com/The-OpenROAD-Project/OpenLane/issues/1925


donn (approved):
lgtm

PR #1931, opened 2023-08-06
Enhancements to variables related to SDC files 

kareefardi:
+ Warn when IMPLEMENTATION_SDC_FILE is not overwritten by the user.
+ Add SDC_IN and IMPLEMENTATION_SDC_FILE_DEFAULT to ignore list in ci Check Variables flow
~ Rename RCX_SDC_FILE to SIGNOFF_SDC_FILE.
~ Rename BASE_SDC_FILE to IMPLEMENTATION_SDC_FILE.
~ Always use IMPLMEMENTATION_SDC_FILE instead of CURRENT_SDC except during signoff stage.


kareefardi:
@donn APU had an increase in the number of antenna violations. I don't see something that this PR does that would affect antenna violations. Perhaps subtle changes in input can affect the output. One solution would be enabling `RUN_HEURISITC_DIODE_INSERTION` and `DIODE_ON_PORTS`. What do you think?

donn:
`+ Warn when IMPLEMENTATION_SDC_FILE is not overwritten by the user.`

Why?

kareefardi:
> `+ Warn when IMPLEMENTATION_SDC_FILE is not overwritten by the user.`
> 
> Why?

It was a request made by @shalan

donn (approved):

PR #1939, opened 2023-08-09
Add validation for @min_distance in io_place.py

kareefardi:
```
+ Add validation for @min_distance in io_place.py
~ Make error and warning messages format of io_place.py consistent with the rest of OL scripts
```


donn:
Folded into #1938

PR #1945, opened 2023-08-16
Create directory store_dir regardless of enable_timestamp value

aoowweenn:
By default, enable_timestamp is True.
Logger handler1 will blame for opening file report_file_name 
 since store_dir is not created priorly.
Adjust the indent of the line of mkdirp to solve the logic error.


donn (approved):

donn:
Thank you so much for the contribution!

PR #1963, opened 2023-08-31
Fix Failing Designs

donn:
This PR's aim is to attempt to fixing failing designs from the daily CI.

I'll start by just fixing a bug identified with OpenLane 2 and playing it by ear from there.

[ci ets]


donn:
{spm, usb} -> "Guide is not connected to design."
y_huff -> Yosys Check Fail
BM64 -> Lint Fail
PicoRV32 -> Failed Heuristic Diode Legalization

PR #1979, opened 2023-09-10
Fix issue where `BASE_SDC_FILE` could not be overridden properly

donn:
~ Fix issue where `PNR_SDC_FILE` and `SIGNOFF_SDC_FILE` were set too early and were not affected by changes to values of `BASE_SDC_FILE` in user config

### DIFF
--- a/scripts/tcl_commands/all.tcl
+++ b/scripts/tcl_commands/all.tcl
@@ -910,8 +910,12 @@ proc prep {args} {
     }
 
 
-    set_if_unset ::env(PNR_SDC_FILE) $::env(BASE_SDC_FILE)
-    set_if_unset ::env(SIGNOFF_SDC_FILE) $::env(BASE_SDC_FILE)
+    if { [info exists ::env(PNR_SDC_FILE) ] } {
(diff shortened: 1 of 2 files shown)


scripts/tcl_commands/all.tcl
```diff
@@ -909,6 +909,14 @@ proc prep {args} {
         }
     }
 
+
+    if { [info exists ::env(PNR_SDC_FILE) ] } {
```
kareefardi, line 913: if not

scripts/tcl_commands/all.tcl
```diff
@@ -909,6 +909,14 @@ proc prep {args} {
...
+        set ::env(PNR_SDC_FILE) $::env(BASE_SDC_FILE)
+    }
+    if { [info exists ::env(SIGNOFF_SDC_FILE) ] } {
```
kareefardi, line 916: if not

kareefardi (changes requested):

kareefardi (approved):

PR #1981, opened 2023-09-10
fix synthesis `chparam` issue

kareefardi:
~ Rename top level module to retain the name after synth params applied


donn (approved):
Confirmed that location in the scripts match OpenLane 2

PR #2002, opened 2023-09-22
Adding the option the use welltap-less PDK

bourquenoud:
Some PDKs don't have welltaps as dedicated cells, typically high voltage technologies with isolation wells. I have slightly modified the scripts such that if no tapcells are defined, it still inserts the decap cells instead of skipping the step altogether.

I am not certain if the `place_endcaps` commands should be placed before or after the `cut_rows` command, but it appears to work for me as it is.

### DIFF
--- a/scripts/openroad/tapcell.tcl
+++ b/scripts/openroad/tapcell.tcl
@@ -14,11 +14,22 @@
 source $::env(SCRIPTS_DIR)/openroad/common/io.tcl
 read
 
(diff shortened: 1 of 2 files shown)


donn (approved):
Thanks for the contribution, will merge when CI passes

donn:
…clicked the wrong button but only like one design was left. ah well :sweat_smile:

PR #2017, opened 2023-10-15
Don't iterate io placement if FP_PIN_ORDER_CFG is defined

kareefardi:


donn (approved):
I'm angry I didn't catch that.

PR #2028, opened 2023-10-25
Update OpenROAD

kareefardi:
~ `openroad_app` -> `0a584d1`

[ci ets]


kareefardi:
@donn would [ci ets] work here

donn (approved):

PR #2046, opened 2023-11-15
Support VERILOG_INCLUDE_DIRS in linter

piotro888:
Adds support for `VERILOG_INCLUDE_DIRS` option to Verilator linter (in `synthesis.tcl` command `run_verilator`)

This option is already supported in synthesis by yosys, so it would be nice if it would also work with linter. Without it, some designs that are synthesized fine, cannot pass the linter.

If preferred, this option could be renamed to `LINTER_INCLUDE_DIRS`, as a independent option from yosys one. I'm not sure which one would fit better.


donn (approved):
looks good- thank you for catching that!

PR #2056, opened 2023-11-21
Issue Survey Fixes

donn:
~ Updated install documentation
~ Fixed issue where the issue survey's PDK checker uses `./pdks` instead of `~/.volare`
- Dropped `pip` requirement/check: `venv`s include `pip` regardless


kareefardi:
> * Dropped `pip` requirement/check: `venv`s include `pip` regardless

Interesting so where does venv get pip from ?

donn:
> > * Dropped `pip` requirement/check: `venv`s include `pip` regardless
> 
> 
> 
> Interesting so where does venv get pip from ? 

As I understand it it packages its own

kareefardi (approved):

Issue #1978, opened 2023-09-09
Example for VSRC_LOC_FILES


kareefardi:
### Description

No example for VSRC_LOC_FILES

### Proposal

Create example for VSRC_LOC_FILES

kareefardi:
Fixed by #2066

PR #2066, opened 2023-12-11
Ameliorate Warnings, PDN Generation, and the default design

donn:
~ Reimplemented the default serial/parallel multiplier to be a bit less cryptic and eliminate linter warnings; and added a proper testbench
~ Reimplemented deprecated variable translation behavior to account for the situation where a deprecated variable's value does not match the default for the new value, which causes a crash
~ Slightly improved warning for designs having been black-boxed during STA
~ PDN Generation Updates
    ~ Renamed `DESIGN_IS_CORE` to `FP_PDN_MULTILAYER` with translation behavior
    ~ PDN generation will now always attempt to extend to core ring (which is inert if core rings do not exist)
    ~ Fixed bug where `FP_PDN_MULTILAYER` being set to `0` would attempt to create a core-ring on two layers anyway
~ IR drop now prints a warning if `VSRC_LOC_FILE` is not provided
- Removed deprecation behavior for `GLB_RT` variables - it's been over a year (>=6 mo as promised)

---

Depends on https://github.com/efabless/openlane-ci-designs/pull/5


scripts/openroad/common/pdn_cfg.tcl
```diff
@@ -108,12 +104,21 @@ if { $::env(FP_PDN_ENABLE_RAILS) == 1 } {
...
+        add_pdn_ring \
+            -grid stdcell_grid \
+            -layers "$::env(FP_PDN_VERTICAL_LAYER) $::env(FP_PDN_HORIZONTAL_LAYER)" \
+            -widths "$::env(FP_PDN_CORE_RING_VWIDTH) $::env(FP_PDN_CORE_RING_HWIDTH)" \
+            -spacings "$::env(FP_PDN_CORE_RING_VSPACING) $::env(FP_PDN_CORE_RING_HSPACING)" \
+            -core_offset "$::env(FP_PDN_CORE_RING_VOFFSET) $::env(FP_PDN_CORE_RING_HOFFSET)"
+    } else {
+        add_pdn_ring \
+            -grid stdcell_grid \
+            -layers "$::env(FP_PDN_VERTICAL_LAYER)" \
```
kareefardi, line 117: Getting this error when entering the else statement
[ERROR PDN-1012] Expecting a list of 2 elements for -layers option of add_pdn_ring command, found 1.
Also doesn't make sense for a ring to have only one layer.

scripts/tcl_commands/sta.tcl
```diff
@@ -79,11 +79,15 @@ proc run_sta {args} {
...
         set file_path [read $fp]
+        set modules [list]
         foreach line [split $file_path "\n"] {
             if { [regexp {module\s+(\S+)\s+not\s+found} $line match first_group] } {
-                puts_warn "Module $first_group blackboxed during sta"
+                lappend modules $first_group
             }
         }
+        if { [llength $modules] > 0 } {
+            puts_warn "The following modules were black-boxed for STA as there was no timing information found: \[[join $modules ", "]\]"
```
kareefardi, line 89: I would prefer replacing ` \[[join $modules ", "]\]` with `\n \[[join $modules "\n\t"]\]`.

scripts/tcl_commands/all.tcl
```diff
@@ -616,64 +652,6 @@ proc prep {args} {
...
-    ### GLB_RT -> GRT (Document using ‡)
-    handle_deprecated_config GLB_RT_ALLOW_CONGESTION GRT_ALLOW_CONGESTION
-    handle_deprecated_config GLB_RT_OVERFLOW_ITERS GRT_OVERFLOW_ITERS
-    handle_deprecated_config GLB_RT_ANT_ITERS GRT_ANT_ITERS
-    handle_deprecated_config GLB_RT_ESTIMATE_PARASITICS GRT_ESTIMATE_PARASITICS
-    handle_deprecated_config GLB_RT_MAX_DIODE_INS_ITERS GRT_MAX_DIODE_INS_ITERS
-    handle_deprecated_config GLB_RT_OBS GRT_OBS
-    handle_deprecated_config GLB_RT_ADJUSTMENT GRT_ADJUSTMENT
-    handle_deprecated_config GLB_RT_MACRO_EXTENSION GRT_MACRO_EXTENSION
-    handle_deprecated_config GLB_RT_LAYER_ADJUSTMENTS GRT_LAYER_ADJUSTMENTS
```
kareefardi, line 671: did you deprecate some variables? i.e. remove `handle_deprecated_config` for some variables.

kareefardi:
Can you also do same as https://github.com/efabless/openlane2-ci-designs/pull/14 here. Should fix https://github.com/The-OpenROAD-Project/OpenLane/issues/1978

scripts/tcl_commands/all.tcl
donn: Yes-- our written policy is >=6 months and the `GLB_RT` variables have been around for about a year, give or take.

scripts/openroad/common/pdn_cfg.tcl
donn: I added an error.

scripts/tcl_commands/sta.tcl
donn: Did something similar-ish.

kareefardi (dismissed):

PR #2072, opened 2023-12-21
Fix a bug where step index was incremented twice before running a step

kareefardi:
Also indicate in the logs if `skip_io` was passed to global placement.


donn (approved):
Looks good, but we should consider removing random global placement in the first place. OpenROAD is now better at handling smaller designs, removing the use case for that.

PR #2073, opened 2023-12-24
Don't warn about blackbox `FILL_CELL`, `DECAP_CELL` and `FP_WELLTAP_CELL` 

kareefardi:
Goal is try to minimize warnings so that users don't miss other important warnings.


donn (approved):
LGTM

PR #2074, opened 2023-12-24
Check if `CLOCK_PORT` is found in the design

kareefardi:
+ Add check for whether the ports listed under `CLOCK_PORT` in fact exist in the top-level module
- Remove misleading CTS log print


donn:
I can do this in Yosys, give me a bit

donn:
Great start

donn (approved):

PR #2109, opened 2024-03-31
Fix Clock Port Check

donn:
~ Clock port check at beginning of flow now allows bits of a bus to be used as clocks
~ Clock port check at beginning of flow no longer allows outputs to be used as clock ports


dlmiles:
A read of the patch looks good to me, can't break it trying to run it as: `python3.9 scripts/check_clock_ports.py  --netlist-in modified.json --top tt_um_module_name "ui_in[0]" "clk" "clk[0]"` if that helps

LGTM

kareefardi (approved):

PR #2115, opened 2024-04-07
Migrate Build System to Match OpenLane 2

donn:
This makes OpenLane use a Nix 3-based build system in place of the ad-hoc `tool_metadata.yaml` + Docker-based build-system.

A Docker image is still the ultimate product, so there is no impact on the end-user (other than `yum` no longer being a thing.)

This PR also effectively synchronizes all tool versions with those of OpenLane 2.

---
```
+ Repository is now a Nix flake
~ Change all invocations of `openroad -python` to use `run_odbpy_script` for consistency
~ Change build system from ad-hoc to Nix, still producing a Docker image as a final result
~ Update KLayout scripts to use `klayout-pymod` or properly parse commandline arguments
~ `open_pdks` -> `bdc9412` to match OpenLane 2
- Remove local installer; `nix run .` will run OpenLane natively
```


docker/build.sh
```diff
@@ -0,0 +1,19 @@
+set -x
+set -e
+TARBALL=$(nix build\
+	--no-link\
+	--print-out-paths\
+	--accept-flake-config\
+	--accept-flake-config\
```
kareefardi, line 7: duplicate line?

donn: mmyes, good catch. will fix with the rest of your inevitable comments

kareefardi (approved):

kareefardi (approved):

PR #2126, opened 2024-05-23
Remove LEC

donn:
- Remove `LEC_ENABLE` and `logic_equiv_check`

They haven't worked properly since 2021 and people keep asking about it so…


kareefardi (changes requested):
Remove `scripts/yosys/logic_equiv_check.tcl`

PR #2134, opened 2024-07-01
Propagate the clock in some STA reports

kareefardi:
+ Add conditional `set_propagated_clocks` for all clocks in `base.sdc`
  ~ Condition is true for:
    ~ CTS STA
    ~ Global routing STA
    ~ Global routing optimizations STA
    ~ Multicorner(signoff) STA
+ Add conditional `set_propagated_clocks` for all clocks after reading an sdc file depending on whether the statement is used in the sdc file for not. Idea is to be backwards compatible with the default behavior of OpenLane of propagating clocks _outside_ the sdc file
+ Add a report for non-propagated clocks
+ Add a report for information regarding clocks


donn (approved):
LGTM

PR #2137, opened 2024-07-09
Increase banner size for OpenLane 2

kareefardi:
~ Increases banner size for OpenLane 2, now uses a PNG rendered at 1200 API on macOS for consistency
~ Updated Flow Architecture Document


donn (changes requested):
Resize is fine by me, PNG is not. Even just on my monitor I can see the pixel aliasing.

donn (approved):

